PR #12, opened 2020-02-14
Auto update packages

icemac:


icemac:
I have to merge this immediately because the TravisCI import feature does not seem to support to define a branch to look for the base file for import.

dataflake:
I have a few more suggestions for `.gitignore`:

These are created when you set up your buildout with `python3 -mvenv` and run builds:

- pyvenv.cfg
- /lib64

These seem to be created on macOS every time I build something with a virtual environment:

- *.profraw

icemac:
See https://github.com/zopefoundation/zLOG/pull/4 for an example PR created by the script.

config/pure-python-without-pypy/tox.ini
```diff
@@ -0,0 +1,40 @@
...
+[testenv:coverage]
+basepython = python3
+deps =
+    coverage
+commands =
+    coverage run -m zope.testrunner --test-path=src []
+    coverage html
+    coverage report
+
+[coverage:run]
```
dataflake: Does this coverage configuration get picked up? I would expect that to be in setup.cfg. If it gets read from here that's fine, I am just wondering.

icemac: Yes, it should according to https://coverage.readthedocs.io/en/v4.5.x/config.html#configuration-files (last paragraph of first section).

config/pure-python-without-pypy/gitignore
```diff
@@ -0,0 +1,14 @@
+*.egg-info
+*.pyc
+*.pyo
```
mgedmin: I suggest also adding `__pycache__/`, although the above rules would probably pick everything under it anyway.

config/pure-python-without-pypy/gitignore
```diff
@@ -0,0 +1,14 @@
...
+.installed.cfg
+.mr.developer.cfg
+.tox/
+bin/
+coverage.xml
+develop-eggs/
+dist/
+htmlcov/
+lib/
+parts/
```
mgedmin: What about `build/`?  And `eggs/`?

config/pure-python-without-pypy/gitignore
icemac: I think `__pycache__` is Python 3 only.

config/pure-python-without-pypy/gitignore
icemac: I'm going to remove `build/` as I do not know where it is needed.
`eggs/` is the default place where eggs are stored by buildout if not using a global eggs directory.

icemac: Sorry, not read correctly. I am going to add `eggs/` but what is `build/` for?

iham: build is created when using zestreleaser or such

dataflake: I think it's created when you build an egg in some situations, like a binary compile?

dataflake: It seems to be created when you run bdist_wheel

iham: most importantly: you can add build/ to gitignores ;) you'll never want it to be added or checked in

icemac: Okay, I am going to add `build/`, too.

PR #25, opened 2020-05-14
Add cli option to prevent direct vcs push.

sallner:
This help during development, debugging and in case of non-trivial changes.

### DIFF
--- a/config/config-package.py
+++ b/config/config-package.py
@@ -25,6 +25,11 @@ parser = argparse.ArgumentParser(
     description='Use configuration for a package.')
 parser.add_argument(
     'path', type=str, help='path to the repository to be configured')
+parser.add_argument(
+    '--no-push',
+    dest='no_push',
+    action='store_true',
+    help='Prevent direct push.')
 parser.add_argument('type', choices=['pure-python',
                                      'pure-python-without-pypy'],
                     help='type of the config to be used, see README.rst')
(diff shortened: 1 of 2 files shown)


jugmac00:
I am not sure whether this should be a command line argument or the default - I have to use the tool myself for some projects.

But now it is merged anyway :-D

sallner:
As the api is not stable yet, this can always be changed. But you might be right.

config/config-package.py
```diff
@@ -78,7 +83,8 @@ def call(*args, capture_output=False):
          'setup.cfg', 'tox.ini', '.gitignore', '.travis.yml', 'MANIFEST.in',
          '.editorconfig')
     call('git', 'ci', '-m', f'Configuring for {config_type}')
```
mgedmin, line 85: So, uh, not related to this PR/commit, but `git ci` is not recognized by git, unless you have an alias in your ~/.gitconfig.  It would be better to use `git commit`.

jugmac00: Good catch!

icemac: Thank you, fixed in 1f7a4a2

jugmac00:
After using the tool once, I really like the tool, but I really dislike the auto-commit and push, as I had to fix stuff afterwards. Just a random internet's guy opinion. Good, you can prevent it.

dataflake:
Another random opinion: I personally **HATE** when a tool makes changes and then commits/pushes immediately. I want changes to be made and then I want to eyeball them before going any further.

mgedmin:
zest.releaser is a tool I trust because

- it shows me the diffs of all the changes it makes and asks for confirmation before committing
- it also asks for confirmation before pushing

OTOH this tool pushes to a branch for a pull request, where I can git commit --amend and git push --force-with-lease to my heart's content, so it's not as bad as it could be.

PR #27, opened 2020-05-14
Fix differences between the two configurations

mgedmin:
I don't see why "without-pypy" implies it won't have Sphinx
documentation, or why the "with-pypy" config is not capable of enforcing
coverage thresholds.  Or why .meta.cfg files should not be ignored by
check-manifest in "with-pypy" configs.

These seem like accidental differences introduced by changing one or the
other config while forgetting the other one exists.

I also snuck in a .gitignore for .coverage.*, because people (and
projects) sometimes use coverage in parallel mode.

### DIFF
--- a/config/pure-python-without-pypy/MANIFEST.in
+++ b/config/pure-python-without-pypy/MANIFEST.in
@@ -7,6 +7,11 @@ include tox.ini
 
 exclude MANIFEST.in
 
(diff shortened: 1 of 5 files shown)


sallner:
Yeah, I was just realizing, that we might need a section/folder with common templates. The main difference should be for travis and tox.

PR #35, opened 2020-09-18
Add support for packages using multi processes in tests like buildout recipes.

icemac:
See zopefoundation/z3c.recipe.tag#8 which was created using this template.

zc.buildout expects an explicit `.coveragerc` so we cannot put the coverage configuration into `tox.ini`.

### DIFF
--- a/config/default/setup.cfg
+++ b/config/default/setup.cfg
@@ -3,6 +3,8 @@ universal = 1
 
 [flake8]
 doctests = 1
+# provided to doctests by buildoutSetUp()
+builtins = write, system, cat, join
 
 [check-manifest]
 ignore =
--- a/config/pure-python-with-multiprocess/packages.txt
+++ b/config/pure-python-with-multiprocess/packages.txt
@@ -1,10 +1,4 @@
 # Packages configured for the pure python version are listed here.
 # Do not edit the file by hand but use the script config-package.py as
 # described in README.rst
-zope.authentication
-z3c.wizard
-zope.browsermenu
-zope.app.publication
-zope.app.broken
-zope.app.appsetup
-zope.browser
+z3c.recipe.tag
--- a/config/pure-python-with-multiprocess/tox.ini
+++ b/config/pure-python-with-multiprocess/tox.ini
@@ -1,0 +1,61 @@
+# Generated from:
+# https://github.com/zopefoundation/meta/tree/master/config/pure-python-with-multiprocess
+[tox]
+envlist =
+    lint,
+    py27,
+    pypy,
+    py35,
+    py36,
+    py37,
+    py38,
+    pypy3,
+    coverage
+
+[testenv]
+usedevelop = true
+deps =
+    zope.testrunner
+commands =
+    zope-testrunner --test-path=src []
+extras = test
+
+[testenv:lint]
+basepython = python3
+skip_install = true
+deps =
+    flake8
+    check-manifest
+    check-python-versions
+commands =
+    flake8 src setup.py
+    check-manifest
+    check-python-versions
+
+[testenv:coverage]
+basepython = python3
+setenv =
+    COVERAGE_PROCESS_START={{toxinidir}}/.coveragerc
+deps =
+    coverage
+    coverage-python-version
+    zope.testrunner
+commands =
+    coverage erase
+    coverage run -m zope.testrunner --test-path=src []
+    coverage combine
+    coverage html
+    coverage report -m {coverage_report_options}
+
+[coverage:report]
+precision = 2
+exclude_lines =
+    pragma: nocover
+    except ImportError:
+    raise NotImplementedError
+    if __name__ == '__main__':
+    self.fail
+    raise AssertionError
+
+[coverage:html]
+directory = htmlcov
--- a/config/README.rst
+++ b/config/README.rst
@@ -19,6 +19,12 @@ packages:
 
   - Configuration for a pure python package which supports PyPy.
 
+* pure-python-with-multiprocess
+
+  - Configuration for a pure python package which supports PyPy and is tested
+    using multiple processes, so coverage has to be configured in a special
+    way. (Used for e. g. buildout recipes.)
+
 * pure-python-without-pypy
 
   - Configuration for a pure python package which does not supports PyPy,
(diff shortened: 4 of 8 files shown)


dataflake:
> zc.buildout expects an explicit `.coveragerc` so we cannot put the coverage configuration into `tox.ini`.

??? where does it say that ???

jamadden:
>> zc.buildout expects an explicit .coveragerc so we cannot put the coverage configuration into tox.ini.

> ??? where does it say that ???

https://github.com/buildout/buildout/blob/cfd54b80966d2b4d7937e3753ba2733cb1c47339/src/zc/buildout/testing.txt#L123

You have to have a file acceptable for the value of `$COVERAGE_PROCESS_START`. My reading of https://coverage.readthedocs.io/en/v4.5.x/config.html#config is that only a `.coveragerc` file is acceptable for that (the special handling of section prefixes only applies when the file name is guessed); but I haven't specifically tested that.

dataflake:
> https://github.com/buildout/buildout/blob/cfd54b80966d2b4d7937e3753ba2733cb1c47339/src/zc/buildout/testing.txt#L123

What I am reading is that this restriction is only required by a special API for testing zc.buildout recipes. There is no such requirement for any "normal" package as far as I can tell.

In my opinion the usage of `.coveragerc` is a bad pattern. It is a dot file, so people overlook it constantly or don't realize it's there at all. It leads to confusion. IMHO the standard should be to use `setup.cfg` wherever possible.

icemac:
> > zc.buildout expects an explicit `.coveragerc` so we cannot put the coverage configuration into `tox.ini`.
> 
> ??? where does it say that ???

When trying to move the coverage configuration to `tox.ini` (as done in the other cases) I get the following error for  zopefoundation/z3c.recipe.tag#8 when calling `tox -ecoverage`:

```
Error in test doctest_tags_recipe (z3c.recipe.tag.tests)
Traceback (most recent call last):
  File "/.../lib/python3.7/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/.../lib/python3.7/unittest/case.py", line 624, in run
    self.setUp()
  File "/.../lib/python3.7/doctest.py", line 2172, in setUp
    self._dt_setUp(test)
  File "/.../z3c.recipe.tag/src/z3c/recipe/tag/tests.py", line 51, in setUp
    zc.buildout.testing.buildoutSetUp(test)
  File "/.../z3c.recipe.tag/.tox/coverage/lib/python3.7/site-packages/zc/buildout/testing.py", line 307, in buildoutSetUp
    assert os.path.isfile(path_to_coveragerc), path_to_coveragerc
AssertionError: /.../z3c.recipe.tag/.coveragerc
```

Maybe I did something wrong. (Yes the restriction for `.coveragerc` only seems to be needed for testing buildout recipes. — Maybe we could rename the configuration template to reflect this.)

config/default/setup.cfg
```diff
@@ -3,6 +3,8 @@ universal = 1
 
 [flake8]
 doctests = 1
+# provided to doctests by buildoutSetUp()
+builtins = write, system, cat, join
```
mgedmin, line 7: Should this be in `default/setup.cfg`?

Very few of ZF packages are buildout recipes that use doctests with `buildoutSetUp()`.

config/pure-python-with-multiprocess/packages.txt
```diff
@@ -0,0 +1,4 @@
+# Packages configured for the pure python version are listed here.
```
mgedmin, line 1: "for the pure python version" *with multiprocessing*, perhaps?

config/pure-python-with-multiprocess/tox.ini
```diff
@@ -0,0 +1,61 @@
...
+    coverage-python-version
+    zope.testrunner
+commands =
+    coverage erase
+    coverage run -m zope.testrunner --test-path=src []
+    coverage combine
+    coverage html
+    coverage report -m {coverage_report_options}
+
+[coverage:report]
```
mgedmin, line 50: This should be in .coveragerc.

config/pure-python-with-multiprocess/tox.ini
```diff
@@ -0,0 +1,61 @@
...
+precision = 2
+exclude_lines =
+    pragma: nocover
+    except ImportError:
+    raise NotImplementedError
+    if __name__ == '__main__':
+    self.fail
+    raise AssertionError
+
+[coverage:html]
```
mgedmin, line 60: This also should be in .coveragerc.

mgedmin (changes requested):
My reading of [coverage.py's documentation](https://coverage.readthedocs.io/en/latest/config.html) suggests that if you have a .coveragerc, then you cannot putt additional coverage configuration in tox.ini, as it will not be used.

config/pure-python-with-multiprocess/tox.ini
icemac: I've overlooked this section, moved it now to `.coveragerc`.

config/pure-python-with-multiprocess/tox.ini
icemac: Same here.

icemac:
BTW putting the coverage configuration to `setup.cfg` gives the same result as using `tox.ini`.

config/default/setup.cfg
icemac: According to https://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations the configuration of `flake8` can be put into `tox.ini`. So I'd like to put it there as well as the `doctests = 1`, so the whole configuration is in one place. (I'd even change the other templates.) Would this be okay for you?

mgedmin: I don't much care where the flake8 configuration lives.  I care about arbitrary getting names excluded from flake8 checks for packages that are not buildout recipes.

mgedmin (approved):

config/default/setup.cfg
icemac: I moved it to `tox.ini`, so the builtins are only defined for buildout-recipes.

config/pure-python-with-multiprocess/packages.txt
icemac: I renamed it to `buildout recipe`.

PR #40, opened 2020-11-08
Quote Python version numbers

jugmac00:
... as with 3.10 this would break anyway.

### DIFF
--- a/config/pure-python/travis.yml
+++ b/config/pure-python/travis.yml
@@ -3,13 +3,13 @@
 language: python
 dist: focal
 python:
-  - 2.7
+  - "2.7"
   - pypy2
-  - 3.5
-  - 3.6
-  - 3.7
-  - 3.8
-  - 3.9
+  - "3.5"
+  - "3.6"
+  - "3.7"
+  - "3.8"
+  - "3.9"
   - pypy3
 
 matrix:
--- a/config/pure-python/travis.yml
+++ b/config/pure-python/travis.yml
@@ -15,10 +15,10 @@ python:
 matrix:
   include:
     - name: "lint"
-      python: 3.7
+      python: "3.7"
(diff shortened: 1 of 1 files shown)


mgedmin (commented):
Seems a bit inconsistent to do it here but not in the matrix.include bits.

mgedmin:
I've released check-python-versions 0.16.1 that will preserve these quotes if you use --add/--drop to update the list of supported version numbers and .travis.yml had all the version numbers quoted.

(check-python-versions was already adding quotes around the special case of "3.10", so it wouldn't be treated as a float.)

icemac:
For me this only adds visual noise.
Why do you think it would break on 3.10? `3.9-dev` also works without quotes.

jugmac00:
`3.9-dev` is a string, `3.10` is a float

mauritsvanrees:
I have not tried what happens, but I can imagine that this will get fixed in tox and/or Travis.
Just a guess.

mgedmin:
This issue has nothing to do with tox.

While Travis could hardcode a special-case for converting Python 3.1 to 3.10 (assuming nobody will ever want the actual Python 3.1), somehow I don't think they will do that.

The right thing to do is to specify Python 3.10 as `"3.10"` in YAML files.

I'm ambivalent on whether we want to quote all other Python versions.  Pros:
- these versions are not floats and we shouldn't be using float syntax
- when people come in to add 3.10 they will be gently directed to follow the example of existing quoted versions and they will avoid a mistake
- consistency

Cons:
- visual noise
- it's not really necessary before 3.10 final is released and supported by Travis (which takes like a month after the upstream release)

dataflake:
My vote would be for consistency, even if it adds visual noise.

icemac (approved):
Okay, convinced.

PR #42, opened 2020-11-18
Unify pure-python templates

icemac:
This is done by adding a `--with-pypy` switch on the command line.
`.meta.cfg` stores whether PyPy is supported or not.


mgedmin (approved):
LGTM

mgedmin:
(It would be nice if we could use Jinja2 instead of %-formatting, but not now.)

icemac:
> (It would be nice if we could use Jinja2 instead of %-formatting, but not now.)

This could be an idea. %-formatting was the only one within Python which does not interfere with the format inside the templates thus did not require any additional quoting.

icemac:
Thank you for looking into this PR. 😀

PR #43, opened 2020-11-20
Switch to Jinja2 as templating engine.

icemac:
Previously we used %-formating.
I kept `%(` and `)s` as start resp. end of variables as the defaults of Jinja2 interfere with the format used in the GitHub actions configuration file.


config/README.rst
```diff
@@ -45,7 +46,7 @@ Each directory contains the following files:
 
   - This file is copied to `.gitignore`.
 
-* MANIFEST.in
+* MANIFEST.in.jj2
```
mgedmin, line 49: This is the first time I see `.jj2`.  I've seen `.j2` before, for Jinja templates.

mgedmin (approved):
LGTM

The %(...)s thing takes some getting used to, but I understand the rationale.

config/README.rst
icemac: I am going to fix it. I've rarely used Jinja2 templates before.

icemac:
Thank you for reviewing this PR. 😀

PR #45, opened 2020-11-24
Add an option to drop Python 2 support.

icemac:
See https://github.com/zopefoundation/importchecker/pull/5 as an example where this is needed.


config/README.rst
```diff
@@ -162,6 +169,9 @@ fail-under
 with-pypy
   Does the package support PyPy: True/False
 
+without-py2
+  Do not run the tests on Python 2, PyPy2 and Python 3.5: True/False
+
```
mgedmin, line 174: Double negatives are hard to parse.  Could we have `with-py2` in the config file, defaulting to True, but keep the `--without-py2` command-line option?

config/README.rst
```diff
@@ -116,6 +116,11 @@ The following arguments are supported.
 --with-pypy
   Enable PyPy support. (Only needed one time as it is stored in .meta.cfg.)
 
+--without-py2
+  The package does not support Python 2, so do not test for it. Also drops
+  support for PyPy2 and Python 3.5 (Only needed one time as it is stored in
+  .meta.cfg.)
```
mgedmin, line 122: I think I sort of understand why 3.5 is bundled here, but it cries for an explanation in the docs.

mgedmin (approved):
I would be a happier person if you accepted my suggested refactoring to avoid double negatives, but I won't veto the change if you don't want to do that.

sallner:
I sketched [here](https://github.com/gocept/meta/pull/1) a possible solution to specify a minimum version and then support all version above that. This would also help with future python versions and projects which do not have Python 3.6 support as they were ported later. This would occasionally save some test builds.

config/README.rst
icemac: I changed the command line option to `--without-legacy-python`. In `.meta.cfg` the parameter is now called `with-legacy-python`. The default is to keep supporting Python 2.7+.

config/README.rst
icemac: As the option is now called `--without-legacy-python` it should be much clearer.

icemac:
@sallner wrote:
> I sketched [here](https://github.com/gocept/meta/pull/1) a possible solution to specify a minimum version and then support all version above that. This would also help with future python versions and projects which do not have Python 3.6 support as they were ported later. This would occasionally save some test builds.

I looked at your PR but decided to keep it here more lightweight as the majority of the repositories here at zopefoundation should run on the same Python versions.

PR #48, opened 2020-12-03
Switch to toml as configuration format.

icemac:
This allows to have comments and indentation in options. (Needed for flake8 additional config.)

See https://github.com/zopefoundation/cipher.lazydate/pull/1 as an example built on the changes in this PR.

Open issues:

- [x] add a converter for the existing .meta.cfg files to port them to the new structure in the toml file. (db13c0d)
- [x] add a script to run the converter script or `config-package.py` on all packages listed in a `packages.txt`


config/README.rst
```diff
@@ -142,21 +142,41 @@ updated. Example:
...
-    additional-flake8-config =
-      ignore = D203
-    additional-check-manifest-ignores =
 
+    [flake8]
+    additional-config = [
+        "# E221 multiple spaces before operator",
+        "# E222 multiple spaces after operator",
+        "per-file-ignores =",
+        "    src/foo/bar.py: E221 E222"
```
mgedmin, line 162: This line is missing a trailing `,`.

config/README.rst
```diff
@@ -142,21 +142,41 @@ updated. Example:
...
+    additional-rules = [
+        "include *.foo",
+        "include *.bar",
+        ]
+
+    [check-manifest]
+    additional-ignores = [
+        "docs/html/*",
+        "docs/source/_static/*",
+        ]
```
mgedmin, line 176: The different alignment of the `]` in the three examples offends my inner pedant.

config/config-package.py
```diff
@@ -40,6 +41,30 @@ def copy_with_meta(template_name, destination, config_type, **kw):
...
+            s = []
+            for u in t:
+                if isinstance(u, list):
+                    for r in u:
+                        s.append(r)
+                else:
+                    retval += u + ",\n    "  # changed from original
+            t = s
+        retval += "]"
+        return retval
```
mgedmin, line 65: If we have to subclass the config file parser/formatter to make it do what we want, why don't we subclass ConfigParser instead to allow leading indentation, like zc.buildout does?

mgedmin (approved):
I'm not a fan of toml, personally, but that's not a good reason to reject this.

LGTM.

config/README.rst
icemac: Fixed.

config/README.rst
icemac: I have now aligned them in the way the file would be written by `config-package.py`.

config/config-package.py
icemac: I only subclassed the encoder to make the result nicer.
Using the default encoder of `toml` dumps lists as:

`foo = [ 'bar', 'baz']`

which is not convenient for longer lists.

Using the base class of my inherited encoder I was able to get it to

```
foo = [ 'bar',
    'baz'
    ]
```

which not completely convinced me. I am going to create a PR for the `toml` package to get my changes upstream.

Subclassing ConfigParser is not as easy as it seems to be. Parsing an ini file is a rather complex process – I did not imagine it being that hard. I tried both `iniparse` and `ConfigUpdater` to work around the problems I had with `ConfigParser`. But both libraries have their individual problems so I decided it might be time to switch to another config file format.

It is not only about keeping the indentation but also about keeping the comments. So they get written to the next config file (e. g. `setup.cfg`).
I am not sure if `zc.buildout` actually keeps the indentation when reading the config file. (Even `ConfigParser` allows to have indentation for multi line values. But it removing the indentation is a feature to prevent values with leading whitespace. If I remember correctly `zc.buildout` does the same.)

icemac:
I tried the `.meta.cfg` to `.meta.toml` converter script in https://github.com/zopefoundation/importchecker/pull/6. It breaks the tests as the repository also needs to be updated with the new linting rules.

icemac:
https://github.com/zopefoundation/importchecker/pull/7 was created using the .meta.cfg to .meta.toml converter in https://github.com/zopefoundation/meta/pull/48/commits/db13c0d85cfe87cebb1d0daa98c8266923fdc505 and it seem to have worked well.

icemac:
https://github.com/zopefoundation/z3c.recipe.sphinxdoc/pull/12 and https://github.com/zopefoundation/zc.recipe.testrunner/pull/9 where created using the new `multi-call.py` script.

icemac:

PR #49, opened 2020-12-07
Ignore .eggs/

mgedmin:
This directory shows up when setuptools has to download some packages itself, usually because of setup_requires.

Because I currently have a log of junk show up in ~/src/zopefoundation/zope.app.publication/.eggs/, and I see that its .gitignore says

```
# Generated from:
# https://github.com/zopefoundation/meta/tree/master/config/pure-python
```


sallner (approved):
That seems like a reasonable addition. :+1:

PR #52, opened 2020-12-18
Add ability to create appveyor.yml.

icemac:
Used in https://github.com/zopefoundation/zope.sendmail/pull/37

Maybe we'll switch to GHA for Windows in the future but for now I just wanted to keep the status quo.

### DIFF
--- a/config/default/MANIFEST.in.j2
+++ b/config/default/MANIFEST.in.j2
@@ -2,6 +2,9 @@ include *.rst
 include *.txt
 include buildout.cfg
 include tox.ini
+{% if with_appveyor %}
+include appveyor.yml
+{% endif %}
 {% if config_type == 'buildout-recipe' %}
 include .coveragerc
 {% endif %}
--- a/config/README.rst
+++ b/config/README.rst
@@ -110,33 +110,36 @@ CLI arguments
 
 The following arguments are supported.
 
---type
-  Define the configuration type (see `Types`_ section above) to be used for the
-  repository. (Only needed one time as it is stored in .meta.toml.)
-
 --no-push
   Avoid pushing at the end of the configuration run.
-
---with-pypy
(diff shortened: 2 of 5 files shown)


config/default/MANIFEST.in.j2
```diff
@@ -2,6 +2,9 @@ include *.rst
 include *.txt
 include buildout.cfg
 include tox.ini
+{% if with_appveyor %}
+include appveyor.yml
+{% endif %}
```
sallner, line 7: Do we need to include the `appveyor.yml` here? We do not include the other CI files up to now, e.g. GHA and others.

sallner (approved):
I just have one comment.

config/default/MANIFEST.in.j2
icemac: According to https://github.com/zopefoundation/zope.sendmail/pull/37/files#diff-41d5a52589e0480be9c099d2bba7a8135b8b0d71bcbb8df3582a8df1c2295003 we previously included at least `.travis.yml`. But `.github/workflows/tests.yml` does not seem to get included without a complaint of `check-manifest`.

sallner: @mgedmin what is you opinion?You have more packaging-foo. I would think it is not needed for distribution.

mgedmin: check-manifest has a default ignore rule of `.github/*`, added back when that directory contained things like issue templates.

I personally prefer to include CI configuration files in my source tarballs for rather silly reasons -- if you're on a desert island with no internet and want to hack on something, and you happen to have a source tarball, then it might be helpful to see what test procedure is being used on various CI systems.

mgedmin (approved):
LGTM

config/default/MANIFEST.in.j2
sallner: But then, we should either put all or none of the ci files into it.

PR #54, opened 2020-12-19
pip install wheel in Appveyor builds

mgedmin:
This appears to be necessary for at least zope.sendmail on Python 3.9,
where (due to lack of binary Windows wheels of zope.proxy for 3.9) pip
install fails to build and install zope.proxy from sources if wheel is
not available.

(I think this is due to a new pip version?  Because older pips used to
fall back to setup.py install if setup.py bdist_wheel failed.)


sallner (approved):
This looks good and is not going to hurt anyhow.

PR #57, opened 2021-01-21
Add support for Zope Product packages.

icemac:
This means we now support a new type which uses `zc.buildout` to install the package under test.

See https://github.com/zopefoundation/Products.MailHost/pull/39 as an example built with this new type.

### DIFF
--- a/config/README.rst
+++ b/config/README.rst
@@ -172,7 +172,6 @@ updated. Example:
 
     [coverage]
     fail-under = 98
-    tox-ini-command = "coverage run {envbindir}/test_with_gs []"
 
     [coverage-run]
     additional-config = [
@@ -192,6 +191,7 @@ updated. Example:
         "setenv =",
         "    ZOPE_INTERFACE_STRICT_IRO=1",
         ]
+    coverage-command = "coverage run {envbindir}/test_with_gs []"
 
     [flake8]
     additional-config = [
@@ -264,12 +264,6 @@ The corresponding section is named: ``[coverage]``.
 fail-under
   A minimal value of code coverage below which a test failure is issued.
 
-tox-ini-command
-  This option replaces the coverage call in the section ``[testenv:coverage]``
-  in ``tox.ini``. *Caution:* only the actual call to collect the coverage data
-  is replaced the reporting calls are not changed. This option has to be a
-  string. If it is not set or empty the default is used.
-
 
 Coverage:run options
 ````````````````````
@@ -296,6 +290,13 @@ testenv-commands
 testenv-additional
   Additional lines for the section ``[testenv]`` in ``tox.ini``.
   This option has to be a list of strings.
+
+coverage-command
+  This option replaces the coverage call in the section ``[testenv:coverage]``
+  in ``tox.ini``. *Caution:* only the actual call to collect the coverage data
+  is replaced. The calls to create the reporting are not changed. This option
+  has to be a string. If it is not set or empty the default is used.
+
 
 Flake8 options
 ``````````````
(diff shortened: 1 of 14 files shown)


config/zope-product/tox.ini.j2
```diff
@@ -0,0 +1,76 @@
...
+
+[testenv:lint]
+basepython = python3
+allowlist_externals =
+    mkdir
+commands_pre =
+    mkdir -p {toxinidir}/parts/flake8
+commands =
+    - isort --check-only --diff {toxinidir}/src setup.py
+    - flake8 --format=html src setup.py
```
sallner, line 34: Do we need the `-` here in front of the command? What about more or less `{toxinidir}`

config/README.rst
```diff
@@ -247,6 +264,12 @@ The corresponding section is named: ``[coverage]``.
 fail-under
   A minimal value of code coverage below which a test failure is issued.
 
+tox-ini-command
```
sallner, line 267: I do not like the wording that much. It makes clear, that it is concerned with `tox.ini` but outside of the coverage section it can mean anything. 
```suggestion
tox-coverage-command
```

sallner (approved):
Some minor comments.

config/zope-product/tox.ini.j2
icemac: It makes that the exit code of the command is ignored. so we see any problems on screen instead only in the HTML file.

config/README.rst
icemac: I was thinking about moving the option to the `[tox]` section (as `coverage-command`) but I am not sure what's the best.

sallner: I wold opt for the tox-section.

config/zope-product/tox.ini.j2
dataflake: I'm guessing most people will use `-p auto` or `-p all` - after finding out about that option and how much time it saves it's become my default.

mgedmin (approved):
lgtm

config/zope-product/tox.ini.j2
icemac: @dataflake So you'd opt to remove the `-` in front of the `isort` call so failures show up even when running the tests in parallel?

config/README.rst
icemac: Fixed.

config/default/tox-coverage-config.j2
```diff
@@ -0,0 +1,22 @@
...
+    pragma: no cover
+    pragma: nocover
+    except ImportError:
+    raise NotImplementedError
+    if __name__ == '__main__':
+    self.fail
+    raise AssertionError
+
+[coverage:html]
+directory = htmlcov
```
dataflake, line 22: This is the one thing I have always disliked about the coverage stuff. It clutters the toplevel folder structure with yet another folder. In all packages I support I usually make it use `parts/coverage` instead.

dataflake (approved):
I haven't used this yet, but it seems to work reasonably well on the two repositories that use it now. My only complaint is about using the toplevel `htmlcov` folder for coverage data as opposed to stuffing that one level deeper into `parts/coverage` - we already use that pattern for linting output that gets stored into `parts/flake8`.

config/default/tox-coverage-config.j2
icemac: I moved `htmlcov` for all types to `parts/htmlcov`, see https://github.com/zopefoundation/meta/pull/57/commits/cba598ba16bcbc037c23fcee642fe2869a818dea

dataflake:
Thank you!

icemac:
Thank you all for reviewing this PR. 😃

PR #60, opened 2021-02-02
Pin the version of `setuptools` until `zc.buildout = 3.0` is released.

icemac:
`easy_install` is not supported anymore by setuptools.


icemac:
Thank you for this PR. 😀

PR #62, opened 2021-02-12
Add option to completely replace the contents of appveyor.yml.

icemac:
This is the only way which makes sense for https://github.com/zopefoundation/Zope/pull/942.
I did not find another way where it does not break.


mgedmin (approved):
lgtm

icemac:

PR #70, opened 2021-03-04
Fix invocation syntax in README

### LINKED ISSUE #67: config tool: unclear how to specify type
on the one hand you can read
`$ bin/python config-package.py <path-to-package> <config-type-name>`

on the other hand you can read
`--type`

So, do I need to use the `--type` option or can I just pass in the type as a positional argument?

mgedmin:
I tried both ways and only one worked.

Fixes #67.

### DIFF
--- a/config/README.rst
+++ b/config/README.rst
@@ -85,7 +85,7 @@ The script needs a ``venv`` with some packages installed::
 
 To use the configuration provided here in a package call the following script::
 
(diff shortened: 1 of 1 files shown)


icemac (approved):
Thank you for caching this. I changed the call signature of the script and forgot to update the documentation, too.

PR #71, opened 2021-03-04
Make it possible to configure [testenv] extras

mgedmin:
Needed for https://github.com/zopefoundation/zope.testrunner/pull/116
where I want to use tox's factors to conditionally enable the 'subunit'
extra for tox environments like py38-subunit.

Apologies for the unrelated flake8 fixes that my editor highlighted to me.

### DIFF
--- a/config/README.rst
+++ b/config/README.rst
@@ -184,6 +184,11 @@ updated. Example:
         "py37-slim",
         "py38-fat",
         ]
+    testenv-extras = [
+        "test",
+        "docs",
+        "extra-feature",
+        ]
     testenv-commands-pre = [
         "{envbindir}/buildout -c ...",
         ]
@@ -318,6 +323,10 @@ additional-envlist
   The configuration for the needed additional environments can be added using
   ``testenv-additional`` (see below). This option has to be a list of strings
   without indentation.
+
+testenv-extras
+  Replacement for the default ``extras`` option in ``[testenv]`` of
+  ``tox.ini``.  This option has to be a list of strings without indentation.
 
 testenv-commands-pre
   Replacement for the default ``commands_pre`` option in ``[testenv]`` of
--- a/config/config-package.py
+++ b/config/config-package.py
@@ -6,9 +6,7 @@ from shared.toml_encoder import TomlArraySeparatorEncoderWithNewline
 import argparse
 import collections
 import jinja2
-import os
 import pathlib
-import shutil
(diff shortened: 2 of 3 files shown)


config/pure-python/packages.txt
```diff
@@ -95,3 +95,4 @@ z3c.pagelet
 zope.app.container
 zope.app.locales
 zope.app.zcmlfiles
+zope.testrunner
```
mgedmin, line 98: oops this probably shouldn't be here yet

icemac:
This breaks the idea to have additional `[testenv]` extras at the top of `testenv-additional`. See example use in https://github.com/zopefoundation/zope.app.locales/pull/13/files#diff-4425ec4892a0640706a7612e3c26160478a52d1cb052428124b2e79769ef72d2R17

I know it is not documented the way I used it. But I had it in mind when coding `testenv-additional` that it could be used for additional extras.
On the other hand I see the benefit of making it explicit.

icemac:
Searching though GitHub it only seems to affect `zope.app.locales` where the PR is not yet merged but no packages which already have been merged, see https://github.com/search?q=org%3Azopefoundation+testenv-additional+language%3ATOML&type=Code&ref=advsearch&l=TOML&l=

config/README.rst
```diff
@@ -319,6 +324,10 @@ additional-envlist
   ``testenv-additional`` (see below). This option has to be a list of strings
   without indentation.
 
+testenv-extras
+  Replacement for the default ``extras`` option in ``[testenv]`` of
```
icemac, line 328: Does it actually have to be a replacement? Having `docs` automatically be added if `--with-sphinx-doctests` is used is actually quite handy.

mgedmin: Eh, I mainly went with a replacement to keep the option name shorter.  I can create a PR for `testenv-additional-extras` if you prefer.

(I'm strongly against packages assuming `testenv-additional` can be used to extend an arbitrary option that happened to be right before the place %(testenv-additional)s got expanded at some point in time.)

icemac: I'd prefer `testenv-additional-extras`.

It was probably not such a good idea to use `testenv-additional` for additional extras in the first place.

mgedmin: I've done that in #76.

icemac: Thank you!

PR #76, opened 2021-03-05
Make it possible to configure [testenv] extras

mgedmin:
Needed for https://github.com/zopefoundation/zope.testrunner/pull/116
where I want to use tox's factors to conditionally enable the 'subunit'
extra for tox environments like py38-subunit.

This is an alternative to #71.

Closes #71.


icemac (approved):
I like it. Thank you.

PR #78, opened 2021-03-08
Look for `tox` on `$PATH` first, then dive into `bin`

### LINKED ISSUE #73: config-package.py shouldn't assume `tox` is in ./bin/
I already have a virtual environment that I use for things like this. I want to use it to run `config-package.py`. The documentation states that it needs a virtual environment with some things installed, and uses an example of making the current directory that virtual environment, but doesn't say that's required. I don't want to create dozens of new virtual environments just to run a tool once.

However, if you don't do that, `config-package.py` fails:
```
Traceback (most recent call last):
  File "../meta/config/config-package.py", line 270, in <module>
    call(pathlib.Path(cwd) / 'bin' / 'tox', '-p', 'auto')
  File "//meta/config/shared/call.py", line 19, in call
    args, capture_output=capture_output, text=True, cwd=cwd)
  File "//lib/python3.7/subprocess.py", line 488, in run
    with Popen(*popenargs, **kwargs) as process:
  File "//lib/python3.7/subprocess.py", line 800, in __init__
    restore_signals, start_new_session)
  File "//lib/python3.7/subprocess.py", line 1551, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: PosixPath('//zope.site/bin/tox'): PosixPath('//zope.site/bin/tox')
```

Creating a symlink to the actual tox in a new bin/ directory worked, but is not ideal.

dataflake:
Fixes #73 

The code will look for a `tox` script on the current `$PATH` and fall back to the current working directory's `bin` subfolder.

### DIFF
--- a/config/README.rst
+++ b/config/README.rst
@@ -98,7 +98,8 @@ The script does the following steps:
 3. Remove a possibly existing ``.coveragerc`` and ``bootstrap.py``. (Coverage
    is now configured in ``tox.ini`` for packages which are no buildout
    recipes.)
-4. Run the tests via: ``tox``
+4. Run the tests via: ``tox``. The ``tox`` script may be either on the current
+   ``$PATH`` or in the ``bin`` subfolder of the current working directory.
 5. Create a branch and a pull request. (Prevent pushing to GitHub using the
    command line switch ``--no-push``.)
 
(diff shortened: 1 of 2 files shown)


mgedmin (approved):
LGTM!

icemac (approved):
LGTM – I tried it locally: it works like a charm.

PR #79, opened 2021-03-08
Add option to prevent automatic commits

### LINKED ISSUE #77: config-package.py: Don't commit by default, or at least have an option not to commit
I want to review the changes before it commits them. Sometimes it takes a few iterations to get it exactly right. I want to be able to `git diff` and see what has been changed, AND I want to set the commit comment.

Scenario: Adding `ZOPE_INTERFACE_STRICT_IRO` in zope.site. It took several tries to discover the right configuration to make that happen, and I ended up with four or five commits, all with an identical comment of "Configuring for pure-python", which of course is not right.

I had to rewrite history to fix this.

dataflake:
Fixes #77 

Adds an option `--no-commit` which prevents the `git add` and `git commit` steps (and implies that nothing will be pushed, either). This makes it easier to inspect the changes or re-run the script without causing more and more commits.

### DIFF
--- a/config/config-package.py
+++ b/config/config-package.py
@@ -32,6 +32,11 @@ parser = argparse.ArgumentParser(
 parser.add_argument(
     'path', type=pathlib.Path, help='path to the repository to be configured')
 parser.add_argument(
+    '--no-commit',
+    dest='no_commit',
+    action='store_true',
+    help='Prevent automatic committing of changes. Implies --no-push.')
+parser.add_argument(
     '--no-push',
     dest='no_push',
     action='store_true',
@@ -297,16 +302,18 @@ with change_dir(path) as cwd:
     else:
         call('git', 'checkout', '-b', branch_name)
         updating = False
-    call('git', 'add',
-         'setup.cfg', 'tox.ini', '.gitignore', '.github/workflows/tests.yml',
-         'MANIFEST.in', '.editorconfig', '.meta.toml')
     if not fail_under:
         print('In .meta.toml in section [coverage] the option "fail-under" is'
               ' 0. Please enter a valid minimum coverage and rerun.')
         abort(1)
-    call('git', 'commit', '-m', f'Configuring for {config_type}')
-    if not args.no_push:
-        call('git', 'push', '--set-upstream', 'origin', branch_name)
+    if not args.no_commit:
+        call('git', 'add',
+             'setup.cfg', 'tox.ini', '.gitignore',
+             '.github/workflows/tests.yml', 'MANIFEST.in', '.editorconfig',
+             '.meta.toml')
+        call('git', 'commit', '-m', f'Configuring for {config_type}')
+        if not args.no_push:
+            call('git', 'push', '--set-upstream', 'origin', branch_name)
     print()
     print('If everything went fine up to here:')
     if updating:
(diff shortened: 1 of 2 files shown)


config/config-package.py
```diff
@@ -31,6 +31,11 @@ def copy_with_meta(template_name, destination, config_type, **kw):
     description='Use configuration for a package.')
 parser.add_argument(
     'path', type=pathlib.Path, help='path to the repository to be configured')
+parser.add_argument(
+    '--no-commit',
+    dest='no_commit',
```
icemac, line 37: We could use the chance to switch away from the negated form when storing the value to a variable as it requires a double negation later on (`if not args.no_commit`). See the `--without-legacy-python` as an example. But the default here should be `true`.

This also applies to `--no-push`.

mgedmin (approved):
LGTM

icemac (commented):
I think it is okay besides one suggestion regarding double negation.

config/config-package.py
dataflake: How do you expect this to work? Do you want to replace `--no-commit` with something like `--commit=False`? That won't work with argparse action `store_false`, it will fail because you cannot pass an argument (`=False`). So you can only say `--commit`, which will do the `store_false` but it makes no sense.

dataflake: Got it after actually looking at `--without-legacy-python`, sorry. Take a look and let me know what you think.

config/config-package.py
```diff
@@ -34,13 +34,13 @@ def copy_with_meta(template_name, destination, config_type, **kw):
     'path', type=pathlib.Path, help='path to the repository to be configured')
 parser.add_argument(
     '--no-commit',
-    dest='no_commit',
-    action='store_true',
+    dest='commit',
+    action='store_false',
```
icemac, line 38: I think it additionally needs `default =True` as its default value is `None`.

icemac (changes requested):
It is much easier to read now. Only an additional default value seems to be needed.

config/config-package.py
```diff
@@ -32,10 +32,15 @@ def copy_with_meta(template_name, destination, config_type, **kw):
...
+    '--no-commit',
+    dest='commit',
+    action='store_false',
+    help='Prevent automatic committing of changes. Implies --no-push.')
 parser.add_argument(
     '--no-push',
-    dest='no_push',
-    action='store_true',
+    dest='push',
+    action='store_false',
```
mgedmin, line 43: This also needs a `default=True`.

mgedmin (changes requested):

config/config-package.py
dataflake: No, default values are created automatically for `store_true` and `store_false`. See https://docs.python.org/3.7/library/argparse.html#action

dataflake: I have added default values to make it easier to understand the code from eyeballing. As I said, they are not needed unless you want to have a default other than the automatically created `True` or `False`.

config/config-package.py
mgedmin: Ooooh, TIL!

mgedmin (approved):

jamadden:
Thank you!

PR #80, opened 2021-03-08
Add option to forego linting

### LINKED ISSUE #75: Suggestion: make flake8 optional
I've an impression that one of the larger roadblocks in GitHub Actions migrations is the requirement for the code base of each package to be flake8-clean.  It requires time and effort from the person who is attempting to create pull requests, as well as from the person who reviews them.

Can we make this optional?  I suggest a `without-flake8 = true` in .meta.toml which would make `tox -e lint` run check-manifest and check-python-versions only.

I can prepare a PR.

dataflake:
Fixes #75 

This PR adds an option `--no-linting` to skip the linting step.

Since linting is part of the `tox` run the solution is a little messy. It involves listing all `tox` environments, removing `lint`,  and then running `tox` with the `TOXENV` environment variable set to the remaining environments

### DIFF
--- a/config/README.rst
+++ b/config/README.rst
@@ -119,6 +119,12 @@ The following arguments are supported.
 --no-push
   Avoid pushing at the end of the configuration run.
 
+--no-linting
+  Don't run ``flake8`` and ``isort`` linters over the code. If the code is old
+  and numerous linting changes would obscure the package changes it may make
+  sense to use this flag and run the linters separate from the package
+  conversion.
+
 --branch
   Define a specific git branch name to be created for the changes. By default
   the script creates one which includes the name of the configuration type.
--- a/config/config-package.py
+++ b/config/config-package.py
@@ -6,6 +6,7 @@ from shared.toml_encoder import TomlArraySeparatorEncoderWithNewline
 import argparse
 import collections
 import jinja2
+import os
 import pathlib
 import toml
 
@@ -36,6 +37,11 @@ parser.add_argument(
     dest='no_push',
     action='store_true',
     help='Prevent direct push.')
+parser.add_argument(
+    '--no-linting',
+    dest='no_linting',
+    action='store_true',
+    help='Do not run the linter over the code.')
 parser.add_argument(
     '--with-appveyor',
     dest='with_appveyor',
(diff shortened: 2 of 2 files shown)


icemac (commented):
My expectation would have been that this option is persisted in the configuration. When using it with the current implementation it will still break on GHA creating a weekly reminder of the broken tests for at least someone (probably the creator of the PR switching to GHA).

But maybe I totally misunderstood the intent of this PR.

dataflake:
You're right, I misunderstood the issue again. He wants a persistent change in `.meta.toml` that will change `tox.ini` and remove the `flake8` run from the `lint` environment.

config/config-package.py
```diff
@@ -286,7 +292,18 @@ def copy_with_meta(template_name, destination, config_type, **kw):
...
+        result = call(pathlib.Path(cwd) / 'bin' / 'tox', '-l',
+                      capture_output=True)
+        tox_envs = result.stdout.split()
+        if 'lint' in tox_envs:
+            tox_envs.remove('lint')
+        os.environ['TOXENV'] = ','.join(tox_envs)
+        call(pathlib.Path(cwd) / 'bin' / 'tox', '-p', 'auto')
+        os.environ['TOXENV'] = old_toxenv or ''
+    else:
+        call(pathlib.Path(cwd) / 'bin' / 'tox', '-p', 'auto')
```
mgedmin, line 306: No, the flag should be passed over to the tox.ini.j2 template and used to skip the flake8/isort steps in [toxenv:lint].

dataflake:
I have changed everything around to omit `flake8` and `isort` from the `tox` `lint` environment and make this a permanent change that is stored in the meta configuration.

config/README.rst
```diff
@@ -148,6 +148,11 @@ The following options are only needed one time as their values re stored in
 --with-sphinx-doctests
   Enable running the documentation as doctest using Sphinx.
 
+--no-flake8
+  Don't add ``flake8`` and ``isort`` linting steps to the configuration. If
```
icemac, line 152: The option does more then the name implies. But I am not able to come up with an alternative name, so I think it is okay.

icemac (approved):
LGTM from reading the code without actually having tried out.

config/README.rst
mgedmin: I like it better than `--no-lint`, given that we still have `tox -e lint` that does some checking.

mgedmin (approved):
LGTM

PR #81, opened 2021-03-10
Run CI on all branch pushes

### LINKED ISSUE #74: config-package.py should run the GHA tests on all pushes, not just to master
A common workflow is to open a branch and iterate on it, pushing it to github to run all the tests on all the interpreters *before* opening a pull request. The PR is only opened once the changes are complete and the CI for the branch is green.

Only running the tests when the PR is opened breaks this. Opening PRs early while still iterating produces lots of useless email notifications for anyone watching the repository.

This completely surprised me when trying to use this tool with zope.site.

dataflake:
Fixes #74 

This change comes closest to the behavior we had before with Travis while avoiding doubling CI checks on pull requests. It runs CI on all branch pushes and omits a separate CI configuration for PRs, which are already fully covered by CI for pushes to the branches they are based on.

### DIFF
--- a/config/default/tests.yml.j2
+++ b/config/default/tests.yml.j2
@@ -3,7 +3,6 @@ name: tests
 on:
   push:
     branches: [ '**' ]
-  pull_request:
   schedule:
     - cron: '0 12 * * 0'  # run once a week on Sunday
 
--- a/config/default/tests.yml.j2
+++ b/config/default/tests.yml.j2
@@ -2,7 +2,7 @@ name: tests
 
 on:
   push:
-    branches: [ master ]
+    branches: [ '**' ]
   pull_request:
   schedule:
     - cron: '0 12 * * 0'  # run once a week on Sunday
--- a/config/default/tests.yml.j2
+++ b/config/default/tests.yml.j2
@@ -3,6 +3,7 @@ name: tests
 on:
   push:
     branches: [ '**' ]
+  pull_request:
   schedule:
     - cron: '0 12 * * 0'  # run once a week on Sunday
 
--- a/config/default/tests.yml.j2
+++ b/config/default/tests.yml.j2
@@ -2,7 +2,6 @@ name: tests
 
 on:
   push:
-    branches: [ '**' ]
   pull_request:
   schedule:
     - cron: '0 12 * * 0'  # run once a week on Sunday


jamadden:
Is this disabling running the tests for PRs, and only running tests on branches? 

My understanding is that running the tests for PRs is considered a good idea, because it's testing something different than the branch tests. The branch tests check that the branch itself passes, but the PR tests that the *merged code* passes. 

Now, if there have been 0 commits to the target branch before merging, those two are (exactly?) the same. But if there have been any commits to the target not included in the branch, the results can be different.

Most often, I think most zopefoundation repos have few open PRs, and the PRs aren't long lived and get merged without other changes to the target branch; and when there have been changes, many contributors rebase or merge (ugh) the target into the working branch. But that's not always the case, and I don't know of a way to require that to be the case, so I'm not sure we want to lose the testing of the merged code.

Importantly, how does this work if the branch is coming from another repository? In that case, would any tests get run (since, if I understand correctly, this doesn't run tests on PRs)?

dataflake:
I've put back the separate PR test. Now the change only does one thing: CI is run on all branches, not just the main branch.

jamadden (approved):
LGTM.

(I haven't used that syntax before, but I trust you. I have just written `on: [push, pull_request]`, but I haven't done anything with scheduled builds.)

dataflake:
Last change, I swear. So omitting any branch specification will work correctly and run CI on the branch and the PR.

The only reason I did not use the abbreviated `on: [push, pull_request]` syntax was that when running `check-python-versions` PyYAML would throw an exception on that syntax.

mgedmin:
> The only reason I did not use the abbreviated on: [push, pull_request] syntax was that when running check-python-versions PyYAML would throw an exception on that syntax.

You cannot do the abbreviated syntax if you also want to have `schedule` in there!

mgedmin (approved):
I was annoyed by seeing four CI statuses (2 * Travis, 2 * Appveyor) instead of two.  I don't think I'll be annoyed by seeing 20 CI statuses, compared to 10: there are too many in any case, and if all of them are green, I see no difference, and if some of them are red, I can play with the little scrollbar.

So the only downside is maybe having to wait a bit longer for CI, and maybe a higher risk of running out of free minutes per month.  And if that happens, we can reconsider.

So, LGTM.

PR #84, opened 2021-03-14
Add isort config for known_local_folder

dataflake:
For `Products.Sessions`, which contains two packages, it is necessary to be able to specify the `isort` `known_local_folder` directive.


mgedmin (approved):
LGTM

PR #85, opened 2021-03-15
Introduce testenv-deps option

mgedmin:
The one and only use-case of this is to make zope.testrunner's own
tox.ini not depend on zope.testrunner.  I made it a configurable list
for maximum flexibility, which might be a mistake.

~~I haven't tested this code yet.~~

I have tested this on zope.testrunner and it works the way I expect.

### DIFF
--- a/config/README.rst
+++ b/config/README.rst
@@ -357,6 +357,11 @@ testenv-commands
   Replacement for the default ``commands`` option in ``[testenv]`` of
   ``tox.ini``. This option has to be a list of strings without indentation.
 
+testenv-deps
+  Replacement for the default ``deps`` option in ``[testenv]`` of ``tox.ini``.
+  This option has to be a list of strings without indentation.  The default is
+  ``['zope.testrunner']``.
+
 testenv-additional
   Additional lines for the section ``[testenv]`` in ``tox.ini``.
   This option has to be a list of strings.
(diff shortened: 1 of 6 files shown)


dataflake:
Might help with packages like AuthEncoding, which for unknown reasons use an entirely different unit test framework...

mgedmin:
> Might help with packages like AuthEncoding, which for unknown reasons use an entirely different unit test framework...

I'm not sure about that, given that [testenv] and [testenv:coverage] commands = have zope-testrunner hardcoded.

dataflake:
> > Might help with packages like AuthEncoding, which for unknown reasons use an entirely different unit test framework...
> 
> I'm not sure about that, given that [testenv] and [testenv:coverage] commands = have zope-testrunner hardcoded.

Hm... never mind then...

icemac (approved):
LGTM.

config/README.rst
```diff
@@ -357,6 +357,11 @@ testenv-commands
   Replacement for the default ``commands`` option in ``[testenv]`` of
   ``tox.ini``. This option has to be a list of strings without indentation.
 
+testenv-deps
+  Replacement for the default ``deps`` option in ``[testenv]`` of ``tox.ini``.
+  This option has to be a list of strings without indentation.  The default is
+  ``['zope.testrunner']``.
```
icemac, line 363: Most of the options are used in the example above, would it be an idea to have this for this option, too?

PR #86, opened 2021-03-18
Prune list of automatically included file types to prevent warnings

dataflake:
In response to https://github.com/zopefoundation/Products.MIMETools/pull/4#discussion_r595822627

This PR cuts down the number of entries in the `MANIFEST.in` template to the absolute minimum. For example, working with the tool has shown that many Sphinx documentation folders don't bother including a Windows `.bat` file for building, and the only thing that is guaranteed to always exist in a `src` folder is Python files.

With this PR, everything else can be added to the `additional-rules` configuration option in the `.meta.toml` `[manifest]` section.

### DIFF
--- a/config/default/MANIFEST.in.j2
+++ b/config/default/MANIFEST.in.j2
@@ -10,18 +10,13 @@ include .coveragerc
 {% endif %}
 {% if with_docs %}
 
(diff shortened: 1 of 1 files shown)


icemac (approved):
LGTM.

PR #93, opened 2021-04-26
Add support for GHA test-commands

icemac:
Plus some other fixed needed for https://github.com/zopefoundation/zope.sqlalchemy/pull/62 resp. fixed while working on that PR.

### DIFF
--- a/config/README.rst
+++ b/config/README.rst
@@ -267,6 +267,7 @@ updated. Example:
         ]
     additional-install = [
         "sudo apt-get update && sudo apt-get install -y libxml2-dev libxslt-dev",
+        "pip install tox-factor"
         ]
     test-commands = [
         "tox -f ${{ matrix.config[1] }}",
--- a/config/README.rst
+++ b/config/README.rst
@@ -367,9 +367,8 @@ testenv-commands
   ``tox.ini``. This option has to be a list of strings without indentation.
 
 testenv-deps
-  Replacement for the default ``deps`` option in ``[testenv]`` of ``tox.ini``.
-  This option has to be a list of strings without indentation.  The default is
-  ``['zope.testrunner']``.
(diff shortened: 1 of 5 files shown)


config/README.rst
```diff
@@ -268,6 +268,9 @@ updated. Example:
     additional-install = [
         "sudo apt-get update && sudo apt-get install -y libxml2-dev libxslt-dev",
         ]
+    test-commands = [
+        "tox -f ${{ matrix.config[1] }}",
```
dataflake, line 272: typo: `-e` instead of `-f`

config/default/setup.cfg.j2
```diff
@@ -44,9 +44,9 @@ ignore-bad-ideas =
 force_single_line = True
 combine_as_imports = True
 sections = FUTURE,STDLIB,THIRDPARTY,ZOPE,FIRSTPARTY,LOCALFOLDER
-known_third_party = %(isort_known_third_party)s
```
dataflake, line 47: Was this intended? No spacing after `=` doesn't look very nice.

dataflake (changes requested):

config/README.rst
icemac: `-f` is intended here, it is an option of https://pypi.org/project/tox-factor/, maybe I should add as an example in the option above.

config/default/setup.cfg.j2
icemac: The space is added in https://github.com/zopefoundation/meta/blob/947b9ba4fbbbae527f79f9c38a4b8c3d86a01b35/config/config-package.py#L177-L184 if needed. Without this change the lines in `setup.cfg` would have trailing whitespace if the options are empty.

dataflake (approved):
Never heard of tox-factor before. Looks like a nice tool.

PR #94, opened 2021-05-05
Add template for packages with c-code.

icemac:
Build on example of zope.interface.

See https://github.com/zopefoundation/zope.interface/pull/243


mgedmin (commented):
lgtm, other than the known issue of using the coveralls action when coverage configuration uses plugins.

icemac:

PR #99, opened 2021-05-20
Ignore htmlcov/

mgedmin:
Several repositories that use pure-python (e.g. zope.browser, zc.queue) leave htmlcov/ directories lying around.

### DIFF
--- a/config/default/gitignore
+++ b/config/default/gitignore
@@ -19,6 +19,7 @@ dist/
 docs/_build
 eggs/
 etc/
+htmlcov/
 lib/
 lib64
 log/


dataflake:
We had instituted a policy - at least for those packages that use buildout - to stick `htmlcov` folders into `parts` and hide them that way.

mgedmin:
It is possible that not all packages have been updated with the newest meta/config, or maybe I have these htmlcov/ directories as relics from older tox runs.

Now I'm wondering if there's any harm from having htmlcov/ in .gitignore.  Hmm, maybe.  If there's a stale htmlcov, then a developer might run `tox -e coverage` and then look at the wrong report.  `git status` might compel them to clean up their working trees and avoid this possibility.

Ok, I retract my PR.

dataflake:
I was pushing to eliminate `htmlcov` at the top level because I find it highly annoying when test/lint/etc processes add stuff there. It's a distraction.

PR #102, opened 2021-06-01
Use package name for coverage source config.

icemac:
This change is needed for https://github.com/zopefoundation/zope.app.publisher/pull/15. Otherwise coverage is 0.
I do not understand why this breaks for this package as all other `pure-python` packages went fine.

### DIFF
--- a/config/default/tox-coverage-config.j2
+++ b/config/default/tox-coverage-config.j2
@@ -2,7 +2,7 @@
 [coverage:run]
 branch = True
 plugins = coverage_python_version
-source = src
+source = %(package_name)s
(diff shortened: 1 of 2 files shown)


dataflake (approved):

config/default/tox-coverage-config.j2
```diff
@@ -2,7 +2,7 @@
 [coverage:run]
 branch = True
 plugins = coverage_python_version
-source = src
+source = %(package_name)s
```
mgedmin, line 5: +:100:

mgedmin (approved):

icemac:

PR #108, opened 2021-06-14
Do not create wheels locally, it has to be done by GHA.

### LINKED ISSUE #107: Upload locally created wheels?
I was releasing version 4.5.1 of  `ExtensionClass` using `zest.releaser`.
Besides the sdist it created `ExtensionClass-4.5.1-cp37-cp37m-macosx_11_0_x86_64.whl` and uploaded it to PyPI.

Is this okay? I was not sure and deleted it immediately – as we have GHA to upload the wheels.
If it is not okay, is there a way to prevent `zest.releaser` from doing so? (Maybe via `create-wheel = false`?)

cc: @jamadden

icemac:
Fixes #107.

### DIFF
--- a/config/default/setup.cfg.j2
+++ b/config/default/setup.cfg.j2
@@ -4,6 +4,11 @@ universal = 1
 {% else %}
 universal = 0
 {% endif %}
+{% if config_type == 'c-code' %}
+
+[zest.releaser]
+create-wheel = no
+{% endif %}
 
 [flake8]
 doctests = 1


dataflake (approved):

config/default/setup.cfg.j2
```diff
@@ -4,6 +4,11 @@ universal = 1
 {% else %}
 universal = 0
 {% endif %}
+{% if config_type == 'c-code' %}
+
+[zest.releaser]
+create-wheel = no
+{% endif %}
```
mgedmin, line 11: +:100:

mgedmin (approved):

icemac:

Issue #97, opened 2021-05-10
RFE: packagize the metaconfiguration script


mgedmin:
Every now and then I want to run the script to update a package, and then I realize that I can't just cd zope.sendmail && run something, I have to juggle two working directories somehow, and then I give up.

I'd like to be able to `pipx install -e ~/src/zopefoundation/meta` and have a runnable `zope-meta` (better name suggestions welcome) script in my PATH.

mgedmin:
TBH a simple wrapper script I could put in ~/bin/ would suffice for me, as long as I can cd ~/src/zopefoundation/something and run zope-meta --help

mgedmin:
So I tried a wrapper like this:

```
#!/bin/sh
# Workaround for https://github.com/zopefoundation/meta/issues/97
config_base=$HOME/src/zopefoundation/meta/config
if ! test -x "$config_base"/bin/python; then
    (cd "$config_base" && python3 -m venv .)
    "$config_base"/bin/pip install -r "$config_base"/requirements.txt
fi
"$config_base"/bin/python "$config_base"/config-package.py "$@"
```

and running `cd ~/src/zopefoundation/zope.interface && zope-meta .` makes config-package.py think the package name is an empty string, with sad results.

icemac:
This will be fixed by #282.

PR #110, opened 2021-06-16
Allow python .../config-package.py .

mgedmin:
See https://github.com/zopefoundation/meta/issues/97#issuecomment-862234270


icemac (approved):
LGTM from reading the code. I did not try it out because I usually run `config-package.py` in its directory with a relative path to the repository it should work on. But I know this is considered suboptimal.

PR #111, opened 2021-06-16
Preserve [tox] flake8

mgedmin:
Currently if you run config-package.py with no arguments, it'll always
set [tox] flake8 = true instead of preserving the existing value.


icemac (approved):
LGTM.

PR #113, opened 2021-06-22
Smarter features

mgedmin:
- When .meta.toml doesn't exist already, autodetect whether the package uses Sphinx and Appveyor
- Make --with-sphinx an alias for --with-docs
- Warn when .meta.toml has Sphinx doctests enabled without having Sphinx docs enabled, because this is likely a mistake by someone passing --with-sphinx when they wanted --with-docs

### DIFF
--- a/config/config-package.py
+++ b/config/config-package.py
@@ -90,6 +90,9 @@ parser.add_argument(
     ' Also disables support for PyPy2.')
 parser.add_argument(
     '--with-docs',
+    # people (me) use --with-sphinx and accidentally get --with-sphinx-doctests
+    # so let's make --with-sphinx an alias for --with-cods
+    '--with-sphinx',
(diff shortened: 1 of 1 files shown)


dataflake (approved):
Good idea, I am sure there are more switches that can be auto-detected.

config/config-package.py
```diff
@@ -90,9 +90,12 @@ def copy_with_meta(template_name, destination, config_type, **kw):
     ' Also disables support for PyPy2.')
 parser.add_argument(
     '--with-docs',
+    # people (me) use --with-sphinx and accidentally get --with-sphinx-doctests
+    # so let's make --with-sphinx an alias for --with-cods
```
mgedmin, line 94: Augh typo

PR #121, opened 2021-08-31
Script to re-enable GitHub Actions.

### LINKED ISSUE #88: GHA auto disable
GitHub disables workflows for repositories which have no activity in 60 days.
See https://docs.github.com/en/actions/managing-workflow-runs/disabling-and-enabling-a-workflow

The documentation states that "scheduled workflows" are disabled but this seems to mean "any workflows" actually.

This happened to me in https://github.com/zopefoundation/grokcore.component/pull/16 where no workflow ran for the first commit.
After looking a while at https://github.com/zopefoundation/grokcore.component/actions I saw a little yellow exclamation mark next to the workflow name. Clicking on it showed up the message that workflows have been disabled.
Just activating workflows was not enough to start the workflow for the PR. A new commit was necessary to start CI tests.

Any idea what we could do about this issue? Or should we be able to live with it?

icemac:
Fixes #88.

I started to use the script but we have to update some of our repositories to get the ability to trigger a workflow manually. (This is already in the the default configuration of `meta/config`, but some repositories where migrated before.)

Open issues:

- [x] use `-R` option instead of cloning the repositories
- [x] if activating fails give the option to clone the repository (so it can easily be updated to the current version of `meta/config`)

### DIFF
--- a/config/README.rst
+++ b/config/README.rst
@@ -633,10 +633,7 @@ Re-enabling GitHub Actions
 After a certain period of time (currently 60 days) without commits GitHub
 automatically disables Actions. They can be re-enabled manually per repository.
 There is a script to do this for all repositories. It does no harm if Actions
-is already enabled.
-
-.. caution:: This script automatically clones all repositories which are
-             configured using the ``config-package.py`` script.
+is already enabled for a repository.
 
 Preparation
 +++++++++++
--- a/config/README.rst
+++ b/config/README.rst
@@ -651,7 +651,6 @@ Preparation
 Usage
 +++++
 
-To run the script point it to a directory where you store clones of the
-zopefoundation repositories::
+To run the script just call it::
 
-    $ bin/python re-enable-actions.py <path-to-clones>
+    $ bin/python re-enable-actions.py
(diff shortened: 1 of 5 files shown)


dataflake (approved):
Just out of curiosity, is it not possible to make these workflow API calls without being in a repository clone? It feels like a drag having to clone everything.

icemac:
@dataflake wrote:
> Just out of curiosity, is it not possible to make these workflow API calls without being in a repository clone? It feels like a drag having to clone everything.

You are right, there is the `-R` option to the GH CLI to work on a repository which is not in the current repository. It even does not need to be cloned at all.

I am going to change the script to use that option.

config/README.rst
```diff
@@ -625,3 +625,32 @@ The script does the following steps:
...
+Re-enabling GitHub Actions
+--------------------------
+
+After a certain period of time (currently 60 days) without commits GitHub
+automatically disables Actions. They can be re-enabled manually per repository.
+There is a script to do this for all repositories. It does no harm if Actions
+is already enabled.
+
+.. caution:: This script automatically clones all repositories which are
+             configured using the ``config-package.py`` script.
```
sallner, line 639: Is this warning still valid?

icemac: Of course not, I am going to fix it.

icemac:
Rebase to current master.

As this PR was already used for many repos to reactivate, it seems to be fine, so I am going to merge it.

PR #125, opened 2021-10-19
Sync Python versions used for lint, docs, coverage between ubuntu and windows.

icemac:


icemac:

PR #126, opened 2021-10-31
Remove Python 3.6 legacy status

dataflake:
This change removes the part of cf84a7fa675ab233c640c6208423a700b8cc874d that moved Python 3.6 to legacy status.

Please see the discussion at https://github.com/zopefoundation/Zope/pull/988#issuecomment-955658477 for why I think it is a bad idea to be aggressive about retiring Python versions.

There's a more general issue here about how this package distinguishes Python versions. I think the distinctions are becoming more of an obstacle for real-life packages which lose Python version support for no good reason, just because a decision was made in this particular package.

Instead of generally thinking about which Python versions are at or near EOL (as this package does) we should think in terms of what our main packages like Zope or ZODB are still supporting. Those must keep good backwards compatibility in a given release series and can't just arbitrarily drop Python version support.


mauritsvanrees (approved):

icemac (approved):
LGTM - I am glad we finally settled on a compromise.

PR #129, opened 2021-11-18
Prevent faulty compiled wheel.

### LINKED ISSUE #128: Faulty compiled wheel on PyPI
Releasing version 4.5.0 of `zope.proxy` resulted in a wheel named `zope.proxy-4.5.0-py2.py3-none-any.whl` on PyPI. This cannot work because the package contains C code which has to be compiled for each and every architecture and Python version. Thus I removed it from PyPI. (https://github.com/zopefoundation/zope.proxy/issues/49 looks like a victim of this faulty wheel.)

According to the PyPI logs that wheel was created by the GHA release process.
The only other 4.5.0 wheel published by the same IP address was `zope.proxy-4.5.0-cp39-cp39-macosx_10_14_x86_64.whl` – maybe this could be a clue.

icemac:
A PyPy never should publish a wheel on PyPI as it cannot be used elsewhere and is not tagged as PyPy only.

Additionally restored that PyPy3 does not run on MacOS. (The version was updated from 3.6 to 3.7 but not the exclude.)

Fixes #128.

### DIFF
--- a/config/c-code/tests-strategy.j2
+++ b/config/c-code/tests-strategy.j2
@@ -30,7 +30,7 @@
 {% endif %}
 {% if with_pypy %}
           - os: macos-latest
-            python-version: "pypy-3.6"
+            python-version: "pypy-3.7"
 {% endif %}
 {% if with_legacy_python %}
           - os: macos-latest
--- a/config/c-code/tests.yml.j2
+++ b/config/c-code/tests.yml.j2
@@ -125,7 +125,7 @@ jobs:
         # We cannot 'uses: pypa/gh-action-pypi-publish@v1.4.1' because
         # that's apparently a container action, and those don't run on
         # the Mac.
-        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags') && startsWith(runner.os, 'Mac')
+        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags') && startsWith(runner.os, 'Mac') && !startsWith(matrix.python-version, 'pypy')
         env:
           TWINE_PASSWORD: ${{ secrets.TWINE_PASSWORD }}
         run: |


jamadden:
> A PyPy never should publish a wheel on PyPI as it cannot be used elsewhere and is not tagged as PyPy only.

I have published from GHA several wheels with C code built by PyPy. They get tagged correctly with names like `RelStorage-3.0.1-pp27-pypy_73-macosx_10_7_x86_64.whl`, `gevent-1.5a1-pp271-pypy_41-macosx_10_14_x86_64.whl`, `gevent-21.8.0-pp27-pypy_73-win32.whl`, and `RelStorage-3.4.5-pp27-pypy_73-macosx_10_7_x86_64.whl `.

If PyPy builds aren't doing that in some projects, more likely there's a deeper configuration issue. Perhaps it's because we completely exclude building any extensions on PyPy for those projects? Without knowing there should be C extensions, the packaging machinery can't know that it's supposed to tag the project. I feel like there's a setting somewhere in setuptools that can be used to override this, but I don't recall off the top of my head.

Or, we could try always building the extensions, even if we don't use them at runtime to force the right tagging.

icemac:
@jamadden Thank you for your comment. I looked a bit into `RelStorage` and `gevent`. The current alpha release of `RelStorage` no longer provides PyPy wheels. Its current final release is too old so the GHA logs are already deleted. `gevent` seems to build its PyPy wheel on AppVeyor.
There are only some PyPy wheels (not for each OS and supported PyPy version).

It is interesting that over there they get tagged correctly. I was not able to find out why it is not the case for `zope.proxy`. But you are probably right, that we do not build the C extension on PyPy at all, see https://github.com/zopefoundation/zope.proxy/blob/083849dcd50f10c5bbab05b55012f04b1c00ccfd/setup.py#L74-L81

To prevent creating faulty wheels for other zopefoundation projects containing C code, I am going to merge this PR to restore the previous behavior where we did not create PyPy wheels. We might find a solution in the future to properly support PyPy.

PR #130, opened 2021-11-19
Run Sphinx doctests only in GHA if the run in tox.

icemac:
This change is only necessary for the c-code template.
The other ones already use tox in GHA.

Needed for https://github.com/zopefoundation/zope.container/pull/43


dataflake (approved):

icemac:

PR #132, opened 2022-01-27
Do not publish future versions

icemac:
Their ABI might not be stable, yet and not compatible across the alpha versions.

Use non-binary coverage version.
They published a wheel not compatible with the current alpha release.
Remove installing coverage in places where it is not needed.

### DIFF
--- a/config/c-code/tests.yml.j2
+++ b/config/c-code/tests.yml.j2
@@ -152,7 +152,13 @@ jobs:
       - name: Install %(package_name)s
         run: |
           pip install -U wheel setuptools
+{% if with_future_python %}
+          # coverage has a wheel on PyPI for a future python version which is
+          # not ABI compatible with the current one, so build it from sdist:
           pip install -U --no-binary :all: coverage
+{% else %}
+          pip install -U coverage
+{% endif %}
           pip install -U 'faulthandler; python_version == "2.7" and platform_python_implementation == "CPython"'
           # Unzip into src/ so that testrunner can find the .so files
           # when we ask it to load tests from that directory. This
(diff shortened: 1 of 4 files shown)


dataflake (approved):

config/c-code/tests.yml.j2
```diff
@@ -130,6 +128,9 @@ jobs:
           && startsWith(github.ref, 'refs/tags')
           && startsWith(runner.os, 'Mac')
           && !startsWith(matrix.python-version, 'pypy')
+{% if with_future_python %}
+          && ${{ matrix.python-version != '%(future_python_version)s' }}
```
mgedmin, line 132: Why are `${{` ... `}}` needed here, but not in other parts of the same conditional expressions that refer to `matrix.python-version`?

config/c-code/tests.yml.j2
```diff
@@ -151,7 +152,7 @@ jobs:
       - name: Install %(package_name)s
         run: |
           pip install -U wheel setuptools
-          pip install -U coverage
+          pip install -U --no-binary :all: coverage
```
mgedmin, line 155: A comment explaining the `--no-binary :all:` would be helpful here so people don't have to dig through git logs.

config/c-code/tests.yml.j2
```diff
@@ -303,6 +304,9 @@ jobs:
         if: >
           github.event_name == 'push'
           && startsWith(github.ref, 'refs/tags')
+{% if with_future_python %}
+          && ${{ matrix.python-version != '%(future_python_version)s' }}
```
mgedmin, line 308: Same syntax question as above.

config/c-code/tests.yml.j2
icemac: I fixed it in a later commit as it did not work as expected, it made the whole expression to become `true`.

config/c-code/tests.yml.j2
icemac: I added a comment in https://github.com/zopefoundation/meta/pull/132/commits/45b1f148eda3f6f46bde852104f8e06c4b3df267

icemac:

PR #133, opened 2022-02-02
Add CONTRIBUTING.md to the root of the target repository

icemac:
The file is a raw copy of in https://github.com/zopefoundation/.github/blob/master/CONTRIBUTING.md
which is not visible enough: It seems only to be referenced in a notice
on the right hand side when creating the first new PR.

See https://github.com/zopefoundation/zodbpickle/pull/64#issuecomment-1026927634 as an example missing the `CONTRIBUTING.md` file.

See https://github.com/zopefoundation/ExtensionClass/pull/41 which was created using this PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,19 @@
+# Contributing to zopefoundation projects
+
+The projects under the zopefoundation GitHub organization are open source and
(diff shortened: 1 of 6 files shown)


dataflake (approved):

icemac:

PR #134, opened 2022-02-08
Add more files and folders to gitignore

wesleybl:


dataflake (approved):

icemac:
Sorry, for being too late to the party, but I do not really like these changes.
The additionally ignored files are no derivates created in the process of running, installing or testing the zopefoundation projects. They are merely the result of a personal workflow or using a specific editor of the users choice. IMHO these files should be globally ignored on the users machine but not in each repository separately.

dataflake:
Thinking about it a little more I think you're right, those don't belong in here. They are artifacts of personal choices regarding tools or habits different from what we normally use. Feel free to revert.

dataflake:
I have reverted the change.

PR #135, opened 2022-03-04
Fix CONTRIBUTING.md and PyPy3 usage:

icemac:
* If CONTRIBUTING.md is not added before the lint tox environment runs it complains that there is a file which is not under version control.
* `pypy3` defaults to `pypy-3.6` which no longer seems to be installed on GHA Windows machines.

Needed for https://github.com/zopefoundation/zope.testrunner/pull/128


dataflake (approved):

icemac:

PR #138, opened 2022-03-09
Update the zope-product template to zc.buildout 3.x.

icemac:
This is needed for https://github.com/zopefoundation/Zope/pull/1014.

### DIFF
--- a/config/zope-product/tox.ini.j2
+++ b/config/zope-product/tox.ini.j2
@@ -3,10 +3,12 @@
 
 [testenv]
 skip_install = true
-# We need to pin setuptools until we have zc.buildout 3.0.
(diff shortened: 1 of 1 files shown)


dataflake:
If this is a safe thing to do then I wonder why there's no final release for zc.buildout 3

dataflake:
If this is a safe thing to do then I wonder why there's no final release for zc.buildout 3

icemac:
We at gocept have been using buildout 3 in production for months.
The problem is the rapidly changing ecosystem (namely `pip` and `setuptools`). Each new version of these tools seems to break `zc.buildout` and there are even fewer developers tackling these issues.

PR #139, opened 2022-03-29
Fix zope-product template:

icemac:
- a current wheel version has to be installed in the first place as otherwise restarting buildout after updating it might fail.
- flake8-html is not compatible with the current version of jinja2 (forcing no flake8 violations make the report obsolete.)

Needed for https://github.com/zopefoundation/Products.MailHost/pull/43


icemac:

PR #140, opened 2022-04-08
Update to zc.buildout 3.0.0rc3.

icemac:
This is the first version which supports a current setuptools version.

Used in https://github.com/zopefoundation/Products.CMFCore/pull/122


icemac:
Drop in coverage is because of the switch to branch coverage.

icemac:

PR #141, opened 2022-04-08
Add isort to linting for type pure-python

dataflake:
This PR adds `isort` to the linting configuration for `pure-python`. It also adds a log file to `.gitignore` that is created by `zope.testing.loggingsupport`, which is used by ZEO tests.

### DIFF
--- a/config/default/gitignore
+++ b/config/default/gitignore
@@ -26,4 +26,5 @@ lib64
 log/
 parts/
 pyvenv.cfg
+testing.log
 var/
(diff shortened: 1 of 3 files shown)


icemac (approved):
I personally do not like the isort rules: At gocept we just sort all imports alphabetically – this prevents from the hassle to find out to which group an import belongs and to configure the groups.

But I see I already lost the battle, so let's try to make imports consistent across more zopefoundation packages.

dataflake:
> But I see I already lost the battle, so let's try to make imports consistent across more zopefoundation packages.

Right, that's my goal here, consistency. It's not about specific rules.

PR #142, opened 2022-04-13
Allow to use pre-release versions when using future python versions.

icemac:
This is needed for https://github.com/zopefoundation/AccessControl/pull/126.

Additionally I had to use `3.11.0-alpha.6` as a7 is not yet on GHA.

### DIFF
--- a/config/config-package.py
+++ b/config/config-package.py
@@ -19,7 +19,7 @@ META_HINT_MARKDOWN = """\
 Generated from:
 https://github.com/zopefoundation/meta/tree/master/config/{config_type}
 --> """
-FUTUTRE_PYTHON_VERSION = "3.11.0-alpha.6"
+FUTURE_PYTHON_VERSION = "3.11.0-alpha.6"
 
 
 def copy_with_meta(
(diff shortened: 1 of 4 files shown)


config/config-package.py
```diff
@@ -19,7 +19,7 @@
 Generated from:
 https://github.com/zopefoundation/meta/tree/master/config/{config_type}
 --> """
-FUTUTRE_PYTHON_VERSION = "3.11.0-alpha.7"
+FUTUTRE_PYTHON_VERSION = "3.11.0-alpha.6"
```
dataflake, line 22:
```suggestion
FUTURE_PYTHON_VERSION = "3.11.0-alpha.6"
```

dataflake (changes requested):
That typo "FUTUTRE" probably needs fixing elsewhere as well.

config/config-package.py
icemac: Thanks for catching this, I am going to fix it in all places.

dataflake (approved):

icemac:

PR #147, opened 2022-05-23
Fix packages which test future Python versions.

icemac:
Some packages testing future Python versions need a RestrictedPython which runs on Python 3.11 so we activated installing pre-released package versions.

Until `repoze.sphinx.autointerface` supports Sphinx 5 we have to use an older version.

Needed by https://github.com/zopefoundation/zope.security/pull/86.


dataflake (approved):

icemac:

PR #150, opened 2022-07-01
Fix building of c-code template for Python 3.10 on MacOS.

icemac:
This is needed for https://github.com/zopefoundation/zope.security/pull/87

### DIFF
--- a/config/c-code/tests.yml.j2
+++ b/config/c-code/tests.yml.j2
@@ -112,7 +112,19 @@ jobs:
           pip install -U pip
           pip install -U setuptools wheel twine cffi
 
-      - name: Build %(package_name)s
+{% for kind in ('Python 3.10 on MacOS', 'all other versions') %}
+      - name: Build %(package_name)s (%(kind)s)
+  {% if kind == 'Python 3.10 on MacOS' %}
+        if: >
+          startsWith(runner.os, 'Mac')
+          && startsWith(matrix.python-version, '3.10')
+        env:
+          _PYTHON_HOST_PLATFORM: macosx-11-x86_64
+  {% else %}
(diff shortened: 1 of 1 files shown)


dataflake:
I don't quite follow. In https://github.com/zopefoundation/zope.security/pull/87 building failed, apparently because of an existing `EGG-INFO` folder. But what does that have to do with this change which restricts macOS to x86_64 only? Building universal eggs is highly desirable.

mauritsvanrees (approved):
Seems worth a try. This environment variable helps when installing on gh-actions.

config/c-code/tests.yml.j2
```diff
@@ -98,21 +98,33 @@ jobs:
...
+{% for kind in ('Python 3.10 on MacOS', 'all other versions') %}
+      - name: Build %(package_name)s (%(kind)s)
+  {% if kind == 'Python 3.10 on MacOS' %}
+        if: >
+          startsWith(runner.os, 'Mac')
+          && startsWith(matrix.python-version, '3.10')
+        env:
+          _PYTHON_HOST_PLATFORM: macosx-11-x86_64
+  {% else %}
+        if: >
```
mauritsvanrees, line 124: If I see it correctly, this condition is not needed: if it is true, nothing happens.

icemac: If it is true the `run` command in line 116+ is executed, see https://github.com/zopefoundation/zope.security/pull/87/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR156-R177 for an example of generated code.

icemac:

PR #152, opened 2022-09-07
Don't run pull_request and push tests simultaneously

dataflake:
This copies a change from the default template to prevent tests for the pull_request event unless the PR is based on an external fork


icemac (approved):
Tested at https://github.com/zopefoundation/ExtensionClass/pull/54: Works like a charm.

PR #153, opened 2022-09-07
Disable unsafe math optimizations

mgedmin:
As @Nephirus wrote on https://github.com/zopefoundation/BTrees/pull/179:

> Current binary wheels use -Ofast optimization option, which (apart from other things) enables flush-to-zero (FTZ) behavior for IEEE-754 subnormals. Unfortunately, FTZ is a global CPU flag, so it affect other running code.

### DIFF
--- a/config/c-code/tests.yml.j2
+++ b/config/c-code/tests.yml.j2
@@ -60,8 +60,8 @@ env:
   PIP_NO_PYTHON_VERSION_WARNING: 1
   PIP_NO_WARN_SCRIPT_LOCATION: 1
 
-  CFLAGS: -Ofast -pipe
-  CXXFLAGS: -Ofast -pipe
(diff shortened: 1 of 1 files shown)


dataflake (approved):

srett:
To whom it may concern (I'm not a Python user): I just stumbled upon [this comment thread on HN](https://news.ycombinator.com/item?id=32740078) that claims `-fno-unsafe-math-optimizations` does **not** override `-Ofast`. Might want to double-check this fix.

mgedmin:
Thank you, @srett!  You gave me the push needed to actually add a regression test and check whether the flags work or not.

And you're absolutely right that it [does not](https://github.com/zopefoundation/BTrees/runs/8266051448?check_suite_focus=true).

Alternative fix (using -O3 instead of -Ofast) is at https://github.com/zopefoundation/meta/pull/155.

PR #154, opened 2022-09-09
Add support for creating *.mo files during the release process.

icemac:
This needs the ability to add `*.mo` to `.gitignore` and the ability to configure `zest.releaser` to require `zest.pocompile`.

It is used in https://github.com/zopefoundation/zope.app.locales/pull/18.


dataflake (approved):

icemac:

PR #155, opened 2022-09-09
Disable unsafe math optimizations (try 2)

mgedmin:
As @Nephirus wrote on https://github.com/zopefoundation/BTrees/pull/179:

> Current binary wheels use -Ofast optimization option, which (apart from other things) enables flush-to-zero (FTZ) behavior for IEEE-754 subnormals. Unfortunately, FTZ is a global CPU flag, so it affect other running code.

Unfortunately -fno-unsafe-math-optimizations does not override -Ofast, so we have to stop using -Ofast instead.

### DIFF
--- a/config/c-code/tests.yml.j2
+++ b/config/c-code/tests.yml.j2
@@ -60,8 +60,8 @@ env:
   PIP_NO_PYTHON_VERSION_WARNING: 1
   PIP_NO_WARN_SCRIPT_LOCATION: 1
 
-  CFLAGS: -Ofast -pipe -fno-unsafe-math-optimizations
-  CXXFLAGS: -Ofast -pipe -fno-unsafe-math-optimizations
+  CFLAGS: -O3 -pipe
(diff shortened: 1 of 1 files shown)


dataflake (approved):

mgedmin:
The following ZF packages will need to be updated to avoid publishing -ffast-math-poisoned wheels to PyPI:
- [ ] AccessControl
- [ ] Acquisition
- [ ] BTrees → https://github.com/zopefoundation/BTrees/pull/184
- [ ] ExtensionClass
- [ ] Persistence
- [ ] persistent
- [ ] zodbpickle
- [ ] zope.container
- [ ] zope.hookable
- [ ] zope.index
- [ ] zope.interface
- [ ] zope.proxy
- [ ] zope.security

icemac:
@mgedmin Thank you very much for providing the list of affected packages: Now each one is fixed and has a new release.

PR #56, opened 2021-01-18
Add ability to configure [testenv] in tox.ini.

icemac:
This is needed for https://github.com/zopefoundation/zope.publisher/pull/58/commits/a897f45ee32214bc09d2d1bdedb75edb97d67de5

PR #157, opened 2022-09-10
Fix path to ``test`` script on Windows

dataflake:
Fixes #56 

This change creates paths to the `test` script using `{envdir}/bin` explicitly instead of `{envbindir}`, which breaks those paths on Windows.

See https://github.com/zopefoundation/DateTime/pull/40/files for a place where I had to fix those paths manually because Windows tests broke.


icemac (approved):
LGTM.

PR #161, opened 2022-10-12
Allow to add additional sources for isort.

icemac:
This PR is needed for https://github.com/zopefoundation/RestrictedPython/pull/234.

### DIFF
--- a/config/README.rst
+++ b/config/README.rst
@@ -284,6 +284,7 @@ updated. Example:
     known_third_party = "ipaddress, PasteDeploy"
     known_zope = "AccessControl, Acquisition, App"
     known_first_party = "Products.GenericSetup, Products.CMFCore"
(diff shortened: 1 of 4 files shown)


dataflake (approved):
I know why this is needed but I still hate the fact that we now have this one package where the tests are in a different place and use a different testing framework.

icemac:
Using `pytest` was an experiment because RestrictedPython 4.0 was a complete rewrite, so we wanted to use the newest tools. The result makes RestrictedPython different form the other packages but at least running the tests via `tox` is the same.

icemac:

dataflake:
Pleas do not forget that any package that uses different tooling or different layout rules represents a high maintenance burden and automatically discourages the usual contributors to help out because all of a sudden they need to learn new tools, just for one package. It's a mistake to change one package's policies just to use the "newest tools".

PR #163, opened 2022-10-27
Add support for Python 3.11 final.

icemac:
The next step would be to declare 3.11 as generally supported version instead of being a future version but there is no 3.12 release on GHA, yet.

Used for https://github.com/zopefoundation/ExtensionClass/pull/56

### DIFF
--- a/config/config-package.py
+++ b/config/config-package.py
@@ -19,7 +19,7 @@ META_HINT_MARKDOWN = """\
 Generated from:
 https://github.com/zopefoundation/meta/tree/master/config/{config_type}
 --> """
-FUTURE_PYTHON_VERSION = "3.11.0-rc.2"
+FUTURE_PYTHON_VERSION = "3.11"
 
 
 def copy_with_meta(
--- a/config/default/appveyor.yml.j2
+++ b/config/default/appveyor.yml.j2
@@ -28,8 +28,8 @@ environment:
 {% if with_future_python %}
     # `multibuild` cannot install non-final versions as they are not on
     # ftp.python.org, so we skip Python 3.11 until its final release:
-    # - python: 311
-    # - python: 311-x64
+    - python: 311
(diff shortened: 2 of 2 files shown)


mgedmin (approved):
lgtm

mgedmin:
> The next step would be to declare 3.11 as generally supported version instead of being a future version but there is no 3.12 release on GHA, yet.

Do we have to wait for 3.12?  Can't we have our templates temporarily ignore it, or produce commented-out 3.12 entries?

dataflake (approved):

icemac:
3.12.0a1 was released two days ago, so I think it will not take too long until it can be used on GHA. Meanwhile we could also run without a future version.

dataflake:
To prevent breakage it makes sense to wait with redefining "future python" until the first 3.12 alpha/beta/whatever is actually available on GHA.

icemac:
As long as RestrictedPython has no final release supporting Python 3.11, we have to keep 3.11 as future version allowing to install non-final releases of dependencies.

So I am now looking into RestrictedPython to push it towards a final release usable for 3.11. After that we can declare 3.12 as new future version and reconfigure the packages currently having `with-future-version = true` to `false` until 3.12 alpha it is available on GHA.

dataflake:
> After that we can declare 3.12 as new future version and reconfigure the packages currently having `with-future-version = true` to `false` until 3.12 alpha it is available on GHA.

Sounds like a lot of manual work - I hope you have that automated somehow.

icemac:
@dataflake wrote:
> Sounds like a lot of manual work - I hope you have that automated somehow.

I think there are only a dozen of packages using `with-future-python = true`. So the additional effort is not too big. Updating a package to support 3.11 is also manual effort. (I think I should document how to do it with at least as possible manual effort.)

PR #164, opened 2022-10-28
Declare Python 3.11 als officially supported.

icemac:
Make Python 3.12 the new future version.
(Even though the first alpha release is not yet available on GHA.)

Maybe we do not have to merge this PR now, but it is needed for https://github.com/zopefoundation/RestrictedPython/pull/243.

### DIFF
--- a/config/c-code/manylinux-install.sh.j2
+++ b/config/c-code/manylinux-install.sh.j2
@@ -34,13 +34,14 @@ tox_env_map() {
         *"cp35"*) echo 'py35';;
 {% endif %}
 {% if with_future_python %}
-        *"cp311"*) echo 'py311';;
+        *"cp312"*) echo 'py312';;
 {% endif %}
         *"cp36"*) echo 'py36';;
         *"cp37"*) echo 'py37';;
         *"cp38"*) echo 'py38';;
         *"cp39"*) echo 'py39';;
         *"cp310"*) echo 'py310';;
+        *"cp311"*) echo 'py311';;
         *) echo 'py';;
     esac
 }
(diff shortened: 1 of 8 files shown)


ale-rt:
Just for reference the Python versions available on GHA are listed here: https://github.com/actions/python-versions/blob/main/versions-manifest.json.

Is it necessary to specify:
```
FUTURE_PYTHON_VERSION = "3.12.0-alpha.1"
```
or can we live with:
```
FUTURE_PYTHON_VERSION = "3.12"
```
?

Otherwise the version will need to be adapted quite frequently.

dataflake (approved):

dataflake:
IMHO we should merge this, otherwise the C-based packages will not upload Python 3.11 eggs to PyPI.

dataflake:
@ale-rt in the docs at https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python#using-a-specific-python-version they say that you can use version specifiers such as `3.x` to get "the latest minor release". There is no mention of development releases, which makes sense. Few people will want to pull development releases.

icemac:
@ale-rt wrote:
> Is it necessary to specify:
> 
> ```
> FUTURE_PYTHON_VERSION = "3.12.0-alpha.1"
> ```

I think it is necessary: I want to test against a specific version of the not next Python version as each alpha, beta or rc release can break things. Yes this is more work but the lessons learned from the 3.11 release are: The changes are quite big between releases and I do not want to sees GHA failures over night because of a new release. (There are already enough of them in supported Python versions.)

icemac:

PR #169, opened 2022-11-09
Drop 32-bit Windows support.

### LINKED ISSUE #165: Appveyor template still tests 32bit Windows
I believe it is a waste of time to run tests on 32bit Windows platforms at this point. Let's remove that and only test on x86_64, please.

icemac:
Not testing 32-bit on Windows also means, we do no longer create 32-bit Windows wheels.

@dataflake Is this the goal you wanted to achieve with #165?

Tested in https://github.com/zopefoundation/ExtensionClass/pull/57.

Fixes #165.

### DIFF
--- a/config/default/appveyor.yml.j2
+++ b/config/default/appveyor.yml.j2
@@ -10,27 +10,18 @@ environment:
 
   matrix:
 {% if with_legacy_python %}
-    - python: 27
     - python: 27-x64
-    - python: 35
     - python: 35-x64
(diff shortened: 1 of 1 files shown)


dataflake (approved):
Yes, my goal is to basically retire Windows 32 bit as a build target for wheels. It has become a very uncommon platform, only used on super low-end hardware nowadays. Servers, which are our target, do not use it anymore.

PR #170, opened 2022-11-09
Update actions versions to fix deprecation warnings

dataflake:
This will get rid of many deprecation warnings about node versions etc.


mgedmin (approved):
lgtm

PR #172, opened 2022-11-10
Add a flag to explicitly add testing on macOS test runners

dataflake:
This works just like `--with-windows` and it allows us to force testing on macOS with packages where it's important, like Zope itself.


mauritsvanrees (approved):
LGTM

icemac (approved):
LGTM.

PR #173, opened 2022-11-11
Enable building x86_64, arm64 and universal2 wheels on supported macOS Python versions

dataflake:
This change applies the same environment flags used by cibuildwheel. Specifically, version 10.9 is explicitly selected when building x86_64 (and universal2) wheels. Only arm64 wheels select version 11.0. See here for details:

https://github.com/pypa/cibuildwheel/blob/2e6cbe2435ef410156319c98625002ed76e08785/cibuildwheel/macos.py#L217-L232

I have tried it on AccessControl master and the wheels get built just fine, see https://github.com/zopefoundation/AccessControl/actions/runs/3445282682/jobs/5748817908

You don't happen to know how to get the equivalent of "python version > 3.10" into a GHA expression, do you? That list of specific versions as condition irks me.

### DIFF
--- a/config/c-code/tests.yml.j2
+++ b/config/c-code/tests.yml.j2
@@ -133,7 +133,9 @@ jobs:
           && (startsWith(matrix.python-version, '3.10')
               || startsWith(matrix.python-version, '3.11'))
         env:
-          _PYTHON_HOST_PLATFORM: macosx-11-x86_64
+          MACOSX_DEPLOYMENT_TARGET: 10.9
+          _PYTHON_HOST_PLATFORM: macosx-10.9-x86_64
+          ARCHFLAGS: -arch x86_64
   {% else %}
           !startsWith(runner.os, 'Mac')
           || !(startsWith(matrix.python-version, '3.10')
--- a/config/c-code/tests.yml.j2
+++ b/config/c-code/tests.yml.j2
@@ -182,6 +182,8 @@ jobs:
         if: matrix.python-version == '%(future_python_version)s'
         run: |
           # Install to collect dependencies into the (pip) cache.
+          # Use "--pre" here because dependencies with support for this future
+          # Python release may only be available as pre-releases
           pip install --pre .[test]
 {% endif %}
       - name: Install %(package_name)s and dependencies
@@ -282,6 +284,8 @@ jobs:
           # when we ask it to load tests from that directory. This
           # might also save some build time?
           unzip -n dist/%(package_name)s-*whl -d src
+          # Use "--pre" here because dependencies with support for this future
+          # Python release may only be available as pre-releases
           pip install --pre -U -e .[test]
 {% endif %}
       - name: Install %(package_name)s
--- a/config/c-code/tests.yml.j2
+++ b/config/c-code/tests.yml.j2
@@ -123,18 +123,6 @@ jobs:
   {% endfor %}
 {% endif %}
 
-{% if with_future_python %}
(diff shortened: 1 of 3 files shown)


dataflake:
The first cut still failed during `BTrees` build on macOS/Python 3.11 as setuptools still got the `persistent` wheel a second time.

I have added a new flag to specify additional packages to install during the `Install Build Dependencies` step. Installing the dependency (in the case of `BTrees` it's `persistent`) at that point will make sure that setuptools knows the egg is already there. It will use that pre-installed egg both for the `setup_requires` and the `install_requires` phase of the setup procedure.

dataflake:
I have now added building x86_64 and arm64 wheels on macOS with Python versions .= 3.8. Python 3.8 is the first version that supports Apple's M-class chips.

arm64 wheels cannot be tested on the GHA runners. So only the x86_64 wheel gets the default name `XXX-${{ runner.os }}-${{ matrix.python-version }}.whl` and is then picked up by the `test` step. The arm64 wheel is uploaded as separate artifact so it can be downloaded when desired.

dataflake:
The first test run of this new configuration is with the BTrees package, see https://github.com/zopefoundation/BTrees/actions/runs/3450155539.

dataflake:
I will merge this tomorrow morning if there's no feedback. I'm traveling later this week and would like to get this done.

mauritsvanrees (approved):
The direction seems good, though I can't follow all the details.

I wonder if it would help for relevant packages to use [PEP 517](https://peps.python.org/pep-0517/) to specify the build system, and then use the [`build`](https://pypa-build.readthedocs.io/en/latest/) tool to create sdist and wheels. But I did not try any of this, I have only heard of it.

config/c-code/tests.yml.j2
```diff
@@ -112,52 +117,128 @@ jobs:
...
       - name: Build %(package_name)s (%(kind)s)
         if: >
-  {% if kind == 'Python 3.10 and 3.11 on MacOS' %}
+  {% if kind == 'macOS x86_64, Python 3.8+' %}
           startsWith(runner.os, 'Mac')
-          && (startsWith(matrix.python-version, '3.10')
-              || startsWith(matrix.python-version, '3.11'))
+          && !(startsWith(matrix.python-version, 'pypy')
+               || matrix.python-version == '2.7'
+               || matrix.python-version == '3.5'
```
icemac, line 136: This looks okay as this list will shrink when we stop supporting Python < 3.6. (Maybe we could even think about dropping Python 3.6 support.)

config/c-code/tests.yml.j2
```diff
@@ -112,52 +117,128 @@ jobs:
...
           python setup.py build_ext -i
           python setup.py bdist_wheel
-          # Also install it, so that we get dependencies in the (pip) cache.
+{% endfor %}
+
+{% if with_future_python %}
+      - name: Install %(package_name)s and dependencies (%(future_python_version)s)
+        if: matrix.python-version == '%(future_python_version)s'
+        run: |
+          # Install to collect dependencies into the (pip) cache.
```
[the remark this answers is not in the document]
dataflake: Good idea, comment added.

config/c-code/tests.yml.j2
dataflake: I wanted to make sure it's a list that doesn't need growing as new Python versions are releases. I'm still unhappy that there's apparently no way to safely express something like `matrix.python-version < '3.8'`, something that understands versioning.

Issue #177, opened 2022-12-19
The time to drop Python 2 support is now!


icemac:
The maintenance burden becomes heavier every week:

* some weeks ago GHA's `ubuntu-latest` no longer contains Python 2.7 up to 3.6, so each repository has to be updated (we are still far away from that update to be complete)
* tox 4 no longer supports the negation syntax used to exclude older python versions from certain actions, see #176
* @vstinner archived `faulthandler` and deleted it from PyPI – we used it for our packages containing C code, to these builds are now also broken, see https://github.com/zopefoundation/BTrees/pull/188 where I try to run without it

We could work around all these issues and the ones to strike in future. But the time it takes, we could perfectly use in other places.
For me the only way to actually make progress is to run https://zope.dev/developer/python2.html#how-to-drop-support on all our repositories before even more problems arise.

I volunteer to rip out Python 2.7 up to 3.6 support form zopefoundation repositories. Even though I know this will take weeks, but fixing the current problems will also take weeks to get GHA green again.

@dataflake @mgedmin @jamadden @d-maurer Any ideas?

dataflake:
Ah, that explains the one failure I saw about `faulthandler`. It boggles my mind why people would just delete releases and willingly break an unknown number of other projects.

In the new release schedule at https://www.zope.dev/releases.html I had already declared Zope 4 as End of Life on 12/31, and Zope 4 was the only reason we have kept Python 2 support going. I am **very** happy to see that support burden go away.

You're right, now is a good time to make the cut. I'm not sure about Python 3.6, though - does it take extra effort or workarounds to keep that in?

mauritsvanrees:
Plone 6.0 requires 3.8+ already.

Python 3.7 is only supported for half a year still, so Zope packages could consider to drop this as well. That version will likely start to give similar problems after a while.

dataflake:
I just remembered that a missing Python 3.6 on `ubuntu-latest` was the reason for the recent test matrix changes, so the question is answered - workarounds are required for Python 3.6 support.

I would suggest that after merging https://github.com/zopefoundation/Zope/pull/1081 I also remove Python 3.6 there and release it as 5.8. We had announced dropping that a while ago, so that feels safe.

d-maurer:
Michael Howitz wrote at 2022-12-19 00:14 -0800:
> ...
>@dataflake @mgedmin @jamadden @d-maurer Any ideas?

I am fine with dropping Python 2 support.

[...]

PR #178, opened 2022-12-19
Drop support for Python 2.7 up to 3.6.

icemac:
See https://github.com/zopefoundation/meta/issues/177

Used in https://github.com/zopefoundation/zope.testing/pull/45

### DIFF
--- a/config/c-code/tox.ini.j2
+++ b/config/c-code/tox.ini.j2
@@ -1,16 +1,13 @@
 {% set with_coverage = True %}
 {% set raw_testenv_setenv = testenv_setenv %}
-{% set testenv_setenv = ['pure: PURE_PYTHON=1', '!pure-!pypy-!pypy3: PURE_PYTHON=0'] + testenv_setenv %}
+{% set testenv_setenv = ['pure: PURE_PYTHON=1', '!pure-!pypy3: PURE_PYTHON=0'] + testenv_setenv %}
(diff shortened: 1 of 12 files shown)


config/c-code/tox.ini.j2
```diff
@@ -1,16 +1,13 @@
 {% set with_coverage = True %}
 {% set raw_testenv_setenv = testenv_setenv %}
-{% set testenv_setenv = ['pure: PURE_PYTHON=1', '!pure-!pypy-!pypy3: PURE_PYTHON=0'] + testenv_setenv %}
+{% set testenv_setenv = ['pure: PURE_PYTHON=1', '!pure-!pypy3: PURE_PYTHON=0'] + testenv_setenv %}
```
dataflake, line 3: Are you sure this double negation works? You can test the result without running the actual tests like this: `bin/tox config -k setenv -epy36`

dataflake (approved):

config/c-code/tox.ini.j2
icemac: I am sure that it will **not** work, but I did not yet have the time ti search for a solution. Currently it seems there is none.

dataflake:
It looks like after initially not caring the tox developers have merged a PR that supposedly fixes the issue. That would require release 4.0.15. I have not tested it.

PR #179, opened 2022-12-20
Add script to drop Python 2.7 up to 3.6 support.

icemac:
Used to create:

* https://github.com/zopefoundation/zope.testbrowser/pull/100

### DIFF
--- a/config/drop-legacy-python.py
+++ b/config/drop-legacy-python.py
@@ -72,7 +72,7 @@ with change_dir(path) as cwd_str:
     print('Replace any remaining code that may support legacy Python 2:')
     call('egrep', '-rn',
          '2.7|3.5|3.6|sys.version|PY2|PY3|Py2|Py3|Python 2|Python 3'
-         '|__unicode__|ImportError"', src, *excludes,
+         '|__unicode__|ImportError', src, *excludes,
          allowed_return_codes=(0, 1))
     wait_for_accept()
     tox_path = shutil.which('tox') or (cwd / 'bin' / 'tox')
--- a/config/drop-legacy-python.py
+++ b/config/drop-legacy-python.py
@@ -63,8 +63,10 @@ with change_dir(path) as cwd_str:
     src = path.resolve() / 'src'
     call('find', src, '-name', '*.py', '-exec',
          bin_dir / 'pyupgrade', '--py3-plus', '--py37-plus', '{}', ';')
+    call(bin_dir / 'pyupgrade', '--py3-plus', '--py37-plus', 'setup.py')
+
     excludes = ('--exclude-dir', '__pycache__', '--exclude-dir', '*.egg-info',
-                '--exclude', '*.pyc')
+                '--exclude', '*.pyc', '--exclude', '*.so')
     print(
         'Replace all remaining `six` mentions or continue if none are listed.')
     call('grep', '-rn', 'six', src, *excludes, allowed_return_codes=(0, 1))
--- a/config/drop-legacy-python.py
+++ b/config/drop-legacy-python.py
@@ -61,8 +61,8 @@ with change_dir(path) as cwd_str:
     print('Remove `six` from the list of dependencies and other Py 2 things.')
     call(os.environ['EDITOR'], 'setup.py')
     src = path.resolve() / 'src'
-    call('find', src, '-name', '*.py',
-         '-exec', 'bin/pyupgrade', '--py3-plus', '--py37-plus', '{}', ';')
+    call('find', src, '-name', '*.py', '-exec',
+         bin_dir / 'pyupgrade', '--py3-plus', '--py37-plus', '{}', ';')
     excludes = ('--exclude-dir', '__pycache__', '--exclude-dir', '*.egg-info',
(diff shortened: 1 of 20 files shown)


config/drop-legacy-python.py
```diff
@@ -0,0 +1,88 @@
...
+    excludes = ('--exclude-dir', '__pycache__', '--exclude-dir', '*.egg-info',
+                '--exclude', '*.pyc')
+    print(
+        'Replace all remaining `six` mentions or continue if none are listed.')
+    call('grep', '-rn', 'six', src, *excludes, allowed_return_codes=(0, 1))
+    wait_for_accept()
+    print('Replace any remaining code that may support legacy Python 2:')
+    call('egrep', '-rn',
+         '2.7|3.5|3.6|sys.version|PY2|PY3|Py2|Py3|Python 2|Python 3'
+         '|__unicode__|ImportError"', src, *excludes,
```
mauritsvanrees, line 75:
```suggestion
         '|__unicode__|ImportError', src, *excludes,
```

With the double quote it looks for a literal `ImportError"`.

config/drop-legacy-python.py
```diff
@@ -0,0 +1,88 @@
...
+        path,
+        f'--branch={branch_name}',
+        '--no-push',
+    ]
+    call(*config_package_args, cwd=cwd_str)
+    print('Remove `six` from the list of dependencies and other Py 2 things.')
+    call(os.environ['EDITOR'], 'setup.py')
+    src = path.resolve() / 'src'
+    call('find', src, '-name', '"*.py"',
+         '-exec', 'bin/pyupgrade', '--py3-plus', '--py37-plus', '{}', ';')
```
mauritsvanrees, line 65: For me on Mac, this does not do anything. But I get no error either, so it is hard to say what is wrong. I guess it is a difference between Mac and Linux.

When I try it myself on the command line I do get an error:

```
$ find /Users/maurits/community/plone-coredev/6.0/src/Products.CMFCore/src -name "*.py" -exec bin/pyupgrade --py3-plus --py37-plus {} ;
find: -exec: no terminating ";" or "+"
```

I tried more quotes, but that did not help. I can never remember or figure out how to use find and exec this way...

I have a `pyupgrade.sh` script that does the following, and that works for me, and does lots of changes after your call has run:

```
git ls-files | grep '\.py$' | xargs pyupgrade $*
```

mauritsvanrees (commented):
This seems good, I started trying it.
But two inline comments: one stray double quote, and the pyupgrade call does not work for me: likely a difference between Mac and Linux that I can't figure out.

config/drop-legacy-python.py
mgedmin: > When I try it myself on the command line I do get an error:

Shell eats the `;` so you need to quote it:

    $ find /Users/maurits/community/plone-coredev/6.0/src/Products.CMFCore/src -name "*.py" -exec bin/pyupgrade --py3-plus --py37-plus {} ";"

> I tried more quotes, but that did not help.

Where did you put those quotes?

> I can never remember or figure out how to use find and exec this way...

I like

    find ... -exec command {} +

because it runs the command once, passing all the filenames as arguments, and you don't need to quote the `+`.

I've no idea if pyupgrade supports multiple filenames on the command line.

config/drop-legacy-python.py
icemac: Good catch.

config/drop-legacy-python.py
icemac: On the command line the `;` has to be quoted. I normally use `\;` so the shell does not interpret it as end of the command. About using `+` I just learnt today.

I am also on a Mac. Probably the `'"*.py"'` has to become `*.py` as there is no shell to interpret the quotes.

mauritsvanrees: I tried quotes at least in these ways I think:

    "pyupgrade --py3-plus --py37-plus" {} ;
    "pyupgrade --py3-plus --py37-plus {}" ;
    "pyupgrade --py3-plus --py37-plus {} ;"

Michael is probably right that the double quotes around `'"*.py"'` were not needed. The latest code on this branch works.

`+` can be good, but every now and then I try to use either `find -exec` or `xargs` in situations where there are too many files, so for example even a simple `ls *` actually fails, say in a buildout downloads cache or shared eggs cache. With `+`, `find` would fail as well, as it would try to fit all arguments on one line. So `";"` is good then.

icemac: Good idea. @mauritsvanrees

dataflake (approved):

dataflake:
I will not be using this script because it gives me no control over what gets committed during its run. At least config_package.py provides ways to say "don't commit anything, I want to control changes manually". One run against Zope right now ended up with changes I don't want, all committed and ready for push so that I deleted the entire clone and started from scratch.

dataflake:
I added a switch for an "interactive" mode that will allow scripts to prompt for input and prevents automatic commits and pushes.

icemac:
It seems time to merge this PR. Thank you for working on it.

PR #183, opened 2022-12-28
Update manylinux docker images for i686 and x86_64 from 2010 to 2014

### LINKED ISSUE #182: manylinux wheel builds for Linux do not build Python 3.11 wheels
It looks like the manylinux builds for Linux - both `i686` and `x68_64` - do not produce wheels for Python 3.11. Those are missing from PyPI and they are missing from the build artifacts of the GHA run, those `manylinux_manylinuxXXXX_ARCH_wheels.zip`. `aarch64` does build Python 3.11 wheels and those are published on PyPI.

dataflake:
Fixes #182 

See test run with `Acquisition` at https://github.com/zopefoundation/Acquisition/actions/runs/3794775048. The logs for the manylinux builds confirm that a Python 3.11 binary egg is built.

### DIFF
--- a/config/c-code/tests.yml.j2
+++ b/config/c-code/tests.yml.j2
@@ -364,7 +364,7 @@ jobs:
     strategy:
       matrix:
         python-version: ["3.9"]
-        image: [manylinux2010_x86_64, manylinux2010_i686, manylinux2014_aarch64]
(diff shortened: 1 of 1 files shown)


gforcada (approved):

icemac (approved):
LGTM, the committed changes seem to be the same as suggested in https://github.com/zopefoundation/persistent/issues/179.

I personally cannot test the result lacking easy access to a linux machine.

PR #185, opened 2022-12-30
Simplify option handling

gforcada:
I'm having a look at the code and I saw a pattern repeated, so I though it can be simplified a little bit, to make it a bit more readable 🍀

### DIFF
--- a/config/config-package.py
+++ b/config/config-package.py
@@ -191,8 +191,11 @@ jinja_env = jinja2.Environment(
 )
 
 
-def set_config_value(name):
-    global meta_cfg
+def set_python_config_value(meta_cfg, args, name, default=False):
+    """Get value from either python section in config file or cmd line arg.
+
+    Return the value and store it in `meta_cfg`.
+    """
     key = f'with-{name}'
     existing_value = meta_cfg['python'].get(key, False)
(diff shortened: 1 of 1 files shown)


icemac (approved):
I like the idea of simplification but added a commit to the PR so the function does not rely on global state and gets a docstring.

config/config-package.py
```diff
@@ -190,26 +190,30 @@ def copy_with_meta(
...
 
+
+def set_python_config_value(meta_cfg, args, name, default=False):
+    """Get value from either python section in config file or cmd line arg.
+
+    Return the value and store it in `meta_cfg`.
+    """
+    key = f'with-{name}'
+    existing_value = meta_cfg['python'].get(key, False)
+    arg_value = vars(args)[f"with_{name.replace('-', '_')}"]
```
mgedmin, line 201:
```suggestion
    arg_value = getattr(args, key.replace('-', '_'))
```

mgedmin (approved):
LGTM, but I also suggest a simplification for a single statement that makes it more readable for me.

gforcada:
Thanks for the review and improvements suggested/applied ✨

PR #187, opened 2023-01-01
refactor: turn it into a class

gforcada:
A bigger step than #185 but on the same direction.

Move as much code as possible to functions.

This way, one can really see the flow/organization of the code much more clear.

The next step, in a future PR would be to create a class to group together all functions and minimize the global usage of variables, which would make it for a much nicer/easier to follow code base IMHO.

Is that something that would be welcomed? 🤔 🍀

### DIFF
--- a/config/config-package.py
+++ b/config/config-package.py
@@ -31,15 +31,16 @@ def copy_with_meta(
 
     If kwargs are given they are used as template arguments.
     """
+    template = jinja_env.get_template(template_name)
+    rendered = template.render(config_type=config_type, **kw)
+    meta_hint = meta_hint.format(config_type=config_type)
+    if rendered.startswith('#!'):
+        she_bang, _, body = rendered.partition('\n')
+        content = '\n'.join([she_bang, meta_hint, body])
+    else:
+        content = '\n'.join([meta_hint, rendered])
+
     with open(destination, 'w') as f_:
-        template = jinja_env.get_template(template_name)
-        rendered = template.render(config_type=config_type, **kw)
-        meta_hint = meta_hint.format(config_type=config_type)
-        if rendered.startswith('#!'):
-            she_bang, _, body = rendered.partition('\n')
-            content = '\n'.join([she_bang, meta_hint, body])
-        else:
-            content = '\n'.join([meta_hint, rendered])
         f_.write(content)
 
 
--- a/config/config-package.py
+++ b/config/config-package.py
@@ -416,6 +416,34 @@ def copy_manifest_in(meta_cfg, path, config_type):
         with_docs=with_docs, with_appveyor=with_appveyor)
 
 
+def copy_appveyor(meta_cfg, path, config_type):
+    appveyor_global_env_vars = meta_cfg['appveyor'].get(
+        'global-env-vars', [])
+    appveyor_additional_matrix = meta_cfg['appveyor'].get(
+        'additional-matrix', [])
+    appveyor_install_steps = meta_cfg['appveyor'].get(
+        'install-steps', ['- pip install -U -e .[test]'])
+    appveyor_build_script = meta_cfg['appveyor'].get(
+        'build-script', [])
+    if config_type == 'c-code' and not appveyor_build_script:
+        appveyor_build_script = ['- python -W ignore setup.py -q bdist_wheel']
+    appveyor_test_steps = meta_cfg['appveyor'].get(
+        'test-steps', ['- zope-testrunner --test-path=src'])
+    appveyor_additional_lines = meta_cfg['appveyor'].get(
+        'additional-lines', [])
+    appveyor_replacement = meta_cfg['appveyor'].get('replacement', [])
+    copy_with_meta(
+        'appveyor.yml.j2', path / 'appveyor.yml', config_type,
+        with_future_python=with_future_python,
+        global_env_vars=appveyor_global_env_vars,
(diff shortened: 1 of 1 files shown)


dataflake:
Making it easier to follow the code is always a good idea. @icemac what do you think?

gforcada:
@dataflake @icemac I went ahead and refactored it once more to use a class, my final goal anyway 🍀 

Now, with `@cached_property` and avoiding that many argument passing at every function, the code looks much nicer IMHO ✨ 

I tried to be careful on copy&pasting the code around and adjusting, but we should double check that nothing got broken.

How does one test for that? 🤔 

If I run it against `zope.interface` from `master` branch and then again from this branch the diff should be almost zero, right? 🤔

config/config-package.py
```diff
@@ -25,473 +26,620 @@
 FUTURE_PYTHON_VERSION = "3.12.0-alpha.2"
```
icemac, line 26:
```suggestion
FUTURE_PYTHON_VERSION = "3.12.0-alpha.2"
DEFAULT = object()
```

config/config-package.py
```diff
@@ -25,473 +26,620 @@
...
+                self.config_type,
+                package_name=self.path.name,
+                setup=manylinux_install_setup,
+                aarch64_tests=manylinux_aarch64_tests,
+                with_future_python=self.with_future_python,
+            )
+            (self.path / '.manylinux-install.sh').chmod(0o755)
+            self.add_manylinux = True
+
+    def cfg_option(self, section, name, default=44):
```
icemac, line 370:
```suggestion
    def cfg_option(self, section, name, default=DEFAULT):
```

config/config-package.py
```diff
@@ -25,473 +26,620 @@
...
+        """An empty list is not a good idea to have it as a default
+
+        Because it is mutable and it gets reused from one call to another.
+        """
+        if default == 44:
```
icemac, line 375:
```suggestion
        if default == DEFAULT:
```

config/config-package.py
```diff
@@ -25,473 +26,620 @@
...
+                with_future_python=self.with_future_python,
+            )
+            (self.path / '.manylinux-install.sh').chmod(0o755)
+            self.add_manylinux = True
+
+    def cfg_option(self, section, name, default=44):
+        """An empty list is not a good idea to have it as a default
+
+        Because it is mutable and it gets reused from one call to another.
+        """
```
icemac, line 374:
```suggestion
        """Read a value from `self.meta_cfg`, default to `[]` if not existing.
        """
```

config/config-package.py
```diff
@@ -25,473 +26,620 @@
...
+        if default == 44:
+            default = []
+        return self.meta_cfg[section].get(name, default)
+
+    def tox_option(self, name, default=44):
```
icemac, line 379:
```suggestion
    def tox_option(self, name, default=DEFAULT):
```

config/config-package.py
```diff
@@ -25,473 +26,620 @@
...
+        """An empty list is not a good idea to have it as a default
+
+        Because it is mutable and it gets reused from one call to another.
+        """
```
icemac, line 383:
```suggestion
        """Read a value from the tox options.
        
        Default not existing ones to `[]`.
        """
```

config/config-package.py
```diff
@@ -25,473 +26,620 @@
...
+        return self.cfg_option('tox', name, default)
+
+    def gh_option(self, name, default=44):
+        """An empty list is not a good idea to have it as a default
+
+        Because it is mutable and it gets reused from one call to another.
+        """
```
icemac, line 390:
```suggestion
    def gh_option(self, name, default=DEFAULT):
        """Read a value from the GitHub actions options.
        
        Default not existing ones to `[]`.
        """
```

config/config-package.py
```diff
@@ -25,473 +26,620 @@
...
+            with_sphinx_doctests=self.with_sphinx_doctests,
+            with_future_python=self.with_future_python,
+            future_python_version=FUTURE_PYTHON_VERSION,
+            with_pypy=self.with_pypy,
+            with_macos=self.with_macos,
+            with_windows=self.with_windows,
+        )
+
+    def manifest_in(self):
+        """Modify MANIFEST.in with meta options"""
```
icemac, line 489:
```suggestion
        """Modify MANIFEST.in with meta options."""
```

icemac (approved):
I like the idea and added some suggestions to refactor the code a bit more.

icemac:
> How does one test for that?

I always use a package where I try my changes and create a PR over there to show what calling that PR in meta/config results in.

dataflake:
You could use some packages that I have configured only yesterday with the latest code from `master`, there should not be any changes except for maybe the `commit-id` value in `.meta.toml`:

- AccessControl
- Acquisition
- BTrees
- zope.security

All of those use the `c-code` template.

gforcada:
@icemac thanks for all the feedback, all applied 👍🏾 

@dataflake thanks for the pointers, I will have a look right away 😃

gforcada:
I did test the few packages that @dataflake suggested, and indeed, only the commit id on `.meta.cfg` was changed ✅ 

This should be ready to get merged 😃

dataflake (approved):

dataflake:
You're free to merge when you have approved reviews, conversations are resolved and tests are green.

PR #188, opened 2023-01-04
Now that Python 2 support is gone we can require tox 4

dataflake:
That way local installs where we have no control over the version of `tox` used can't get too badly out of sync with GitHub Actions installs, which will always grab the latest version.


mgedmin (approved):
Are we actually using any tox-4-only features?  Or is this to make sure people making changes to tox.ini files won't accidentally introduce incompatibilities tox 4?

dataflake:
We are not using tox 4 features but I have been seeing several issues moving from one version to the next. So it's better to force local installs to update so we don't get error reports for issues that have to do with tox incompatibilities.

icemac (approved):

PR #192, opened 2023-02-03
Inherit the `allowlist_externals` to the coverage env.

icemac:
This is needed for zope.sqlachemy.

Used in https://github.com/zopefoundation/zope.sqlalchemy/pull/77


dataflake (approved):
While you're working in the package, Python 3.12 is now available as "3.12.0-alpha.4".

icemac:
@dataflake I'll give Python 3.12 a try this week at Alpine City sprint. I already know that AccessControl does not compile.

PR #197, opened 2023-03-30
Ignore generated translation files

dataflake:
This came up in conjunction with the partial Japanese translation of the Zope docs. We don't want to ship the generated files, just the sources.

### DIFF
--- a/config/default/gitignore.j2
+++ b/config/default/gitignore.j2
@@ -1,5 +1,6 @@
 *.dll
 *.egg-info/
+*.mo
 *.profraw
 *.pyc
 *.pyo


icemac:
> We don't want to ship the generated files, just the sources.

That's generally true but we have [some packages](https://github.com/search?q=org%3Azopefoundation+%22ignore-bad-ideas+%3D%22+NOT+is%3Aarchived&type=code&l=Python) storing there *.mo files in the repository and expecting them to be there. 

Event Zope itself stores some *.mo files, see https://github.com/zopefoundation/Zope/blob/master/setup.cfg#L27-L29

dataflake:
> That's generally true but we have [some packages](https://github.com/search?q=org%3Azopefoundation+%22ignore-bad-ideas+%3D%22+NOT+is%3Aarchived&type=code&l=Python) storing there *.mo files in the repository and expecting them to be there.

Then I suggest parameterization for the `.gitignore` template.

> Event Zope itself stores some *.mo files, see https://github.com/zopefoundation/Zope/blob/master/setup.cfg#L27-L29

Not an argument: Those are test artifacts and not functional documentation translations.

dataflake:
I stand corrected, you already added `.gitignore` parameterization. I'll remove the blanket ignore and will see if I can add a specific one to `.meta.toml` that just ignores them in the docs folder.

PR #202, opened 2023-08-23
Use the `meta` repository to compute the branch.

icemac:


dataflake (approved):
There's no comment, but I am guessing this is just to make extra sure the meta repository commit id is used for cases where someone runs the script outside of the meta clone, so that's fine.

icemac:
Previously the git commit hash of the package to be configured was used now it is using the commit hash of the `meta` repository.

PR #204, opened 2023-09-13
Add toolkit template.

icemac:
Will be used for zopetoolkit and groktoolkit.

The previous mentioning of a groktoolkit template was not true, it never made it out of my hard disk into the repository as it was not ready.

Used in https://github.com/zopefoundation/zopetoolkit/pull/57.


dataflake (approved):

icemac:

PR #205, opened 2023-09-21
add readthedocs support

### LINKED ISSUE #203: Create .readthedocs.yml for packages having documentation
If a package has documentation it is probably rendered on ReadTheDocs.
That service requires some configuration which can be automatically created/updated by `meta/config`.

See https://github.com/zopefoundation/zope.i18n/pull/58/files for a working example.

dhavlik:
Fixes #203


config/config-package.py
```diff
@@ -328,6 +328,12 @@ def gitignore(self):
             ignore=git_ignore,
         )
 
+    def readthedocs(self):
+        self.copy_with_meta(
+            'readthedocs.yaml.j2', self.path / '.readthedocs.yaml',
```
icemac, line 333: A `git add` call is missing for this new file.

icemac (changes requested):

config/config-package.py
dhavlik: added

icemac (approved):
LGTM.

icemac:
> How to use this is not documented in the README

I added https://github.com/zopefoundation/meta/pull/212 to fix this.

PR #207, opened 2023-09-22
Set branch protection rules.

### LINKED ISSUE #191: Automatically set Branch protection rules
Currently I set the "required status checks to pass before merging" by hand:

* docs (if applicable)
* lint
* coverage
* py37 (as oldest supported python version)
* py311 (as newest supported python version)
* pypy3 (if applicable as slowest running test)

Repositories containing C code or testing against Windows or MacOS have other status check names.

The actual list of status check names depends on the configuration via `meta/config` so it should be set when running it.

icemac:
See #191


dataflake:
Does the branch protection script make sure that the "Release Managers" group can override restrictions such as pushing directly to master?

icemac:
> Does the branch protection script make sure that the "Release Managers" group can override restrictions such as pushing directly to master?

No, this is not yet done. But I want to get it merged first and then I can add it.

config/shared/call.py
```diff
@@ -36,5 +37,12 @@ def call(*args, capture_output=False, cwd=None, allowed_return_codes=(0, )):
     result = subprocess.run(
         args, capture_output=capture_output, text=True, cwd=cwd)
     if result.returncode not in allowed_return_codes:
-        abort(result.returncode)
+        if capture_output:
+            abort_text = textwrap.defent(f'''
```
dataflake, line 41: Is this a typo?

```suggestion
            abort_text = textwrap.dedent(f'''
```

dataflake (approved):

dataflake:
With the typo I pointed out fixed this should be merged because I already keep using it due to the (unrelated) changes in GHA runner versions.

config/shared/call.py
icemac: Yes it is a typo. So this code has never been run. Shame on me.
Thank you for spotting it.

icemac:

PR #208, opened 2023-09-22
#112 Add windows to github actions

### LINKED ISSUE #112: Build Windows wheels on GHA
Currently we use AppVeyor. This works okay for now. But it is a slow external CI system.
There is already the ability to run the tests on Windows via GHA (run `meta/config` using `--with-windows`). But it does not yet build binary wheels for windows.

vernans:
Closes #112 

It's not pretty but the built wheels seem to work.

Persistent and ExtensionClass were tested

https://github.com/zopefoundation/ExtensionClass/pull/72

https://github.com/zopefoundation/persistent/pull/190

When updating the repositories, with_appveyor needs to be set to false, also I'd set with_windows to true even though it is not used in c-code packages at the moment.


dataflake (approved):

dataflake:
I have made one change: Instead of always selecting the Windows runners I want the `with_windows` flag to govern that. So setting `with_windows` has a meaning. A quick manual test shows me that it does the right thing.

dataflake:
@icemac I like this idea a lot. If the process of creating wheels and automatically uploading them to PyPI when a tag is encountered works as before then this is much better than Appveyor.

After updating the 13(!) binary packages Zope depends on yesterday my choice is for moving everything to GitHub. For some packages it took more than two hours for the Windows wheels to appear on PyPI. Appveyor behavior and speed was very erratic. The queue seemed to hang forever on some of the projects.

icemac:
@vernans @dataflake I additionally completely removed appveyor and added support for Python 3.13 on Windows, see https://github.com/zopefoundation/ExtensionClass/pull/72 as an example.

Please (re-)review.

config/buildout-recipe/packages.txt
```diff
@@ -9,7 +9,6 @@ z3c.recipe.i18n
 zc.recipe.cmmi
 zdaemon
 zc.recipe.filestorage
-five.customerize
```
dataflake, line 12: Was this intentional? Just wondering.

dataflake (approved):

config/buildout-recipe/packages.txt
icemac: It ended up in this PR by accident, but the change is correct, see https://github.com/zopefoundation/meta/pull/208/commits/f33185d73bbb353c449fc7ed8d0e3499fb0a00d7 – I am moving it to `master` to clean up the PR.

dataflake (approved):

PR #211, opened 2023-10-05
Switch Python 3.12 to non-future and change with-future-python behavior

dataflake:
This PR does two things:

- move Python 3.12 to the list of standard non-future Pythons
- prepare places that want to use future Pythons by using `py313`
- redefine `FUTURE_PYTHON_VERSION` to an empty string
- only render future python-related template items if `FUTURE_PYTHON_VERSION` is not empty

With the last two changes individual packages do not need to turn off their `with-future-python` flag during those times when the Python release is finished and no prerelease exists for the next Python version.

### DIFF
--- a/config/c-code/manylinux-install.sh.j2
+++ b/config/c-code/manylinux-install.sh.j2
@@ -30,13 +30,14 @@ yum -y install libffi-devel
 tox_env_map() {
     case $1 in
 {% if with_future_python %}
-        *"cp312"*) echo 'py312';;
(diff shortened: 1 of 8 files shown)


dataflake:
A manual test on the `DocumentTemplate` repo shows that the only change individual packages will need for "official" Python 3.12 support is adding the trove classifier in `setup.py`.

icemac (approved):
I like the changes.
The first alpha of 3.14 is due to be released in about 2 weeks, so we are prepared.

PR #213, opened 2023-10-05
Add ability to specify additional RTD build options

dataflake:
This adds support for additional lines to be inserted into the `.readthedocs.yaml` build configuration section. This is useful to e.g. install additional packages in the OS.

### DIFF
--- a/config/README.rst
+++ b/config/README.rst
@@ -377,6 +377,12 @@ updated. Example:
         "*.mo",
         ]
 
+    [readthedocs]
+    build-extra = [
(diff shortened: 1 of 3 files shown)


icemac (approved):
LGTM, but where is it needed?

dataflake:
This is needed for things like database adapters that must have database libraries or databases installed on the test server. I use it for Products.LDAPUserFolder right now, but it will also be needed when updating Products.ZpsycoPGDA and Products.ZMySQLDA

PR #214, opened 2023-10-05
Don't call "git add" if the --no-commit flag is set

dataflake:
This change solves an ongoing annoyance for my own workflow: I always use `--no-commit` because I want to eyeball all changes. But several places use `git add`, which puts the changes into the git staging area and thus out of reach of `git diff`. I always have to do `git restore --staged` for those files. With this change `git add` is not done when `--no-commit` is used.


icemac (approved):
LGTM.

PR #209, opened 2023-09-22
Use the toml implementation from standard lib.

sallner:
Unfortunately, the standard lib only supplies read-only functionality. `tomli-w` is a small `TOML` writing library. I should be enough for our purpose, but it cannot write comments

Using `tomlkit` would allow fine more finely grained editing, but comes with a whole new data structure and needs a bigger refactoring.

The changes in this PR will change the formatting of the .meta.toml slightly, see the result for the Zope package

<details>
<summary> Diff of the .meta.toml in Zope </summary>

```diff
diff --git a/.meta.toml b/.meta.toml
index 9a080cba6..c352105f5 100644
--- a/.meta.toml
+++ b/.meta.toml
@@ -2,7 +2,7 @@
 # https://github.com/zopefoundation/meta/tree/master/config/zope-product
 [meta]
 template = "zope-product"
-commit-id = "a3ddb354"
+commit-id = "77efeb4b"
 
 [python]
 with-pypy = false
@@ -26,23 +26,23 @@ known_zope = "AccessControl, Acquisition, App, DateTime, DocumentTemplate, Exten
 additional-config = [
     "# W503 line break before binary operator",
     "ignore = W503",
-    ]
+]
 
 [tox]
 testenv-deps = [
     "cffi",
-    ]
+]
 additional-envlist = [
     "pre-commit",
-    ]
+]
 testenv-commands-pre = [
     "{envbindir}/buildout -c {toxinidir}/buildout.cfg buildout:directory={envdir} buildout:develop={toxinidir} buildout:root-directory={toxinidir}",
-    ]
+]
 testenv-commands = [
     "# the `layer` argument below suppresses a `ZCatalog` performance test",
     "# which occasionally fails on GITHUB",
     "{envdir}/bin/alltests --layer '!Products.PluginIndexes' {posargs:-vc}",
-    ]
+]
 coverage-basepython = "python3.8"
 coverage-command = "coverage run {envdir}/bin/alltests {posargs:-vc}"
 testenv-additional = [
@@ -68,7 +68,7 @@ testenv-additional = [
     "commands =",
     "    autopep8 --verbose --in-place --recursive --aggressive --aggressive {toxinidir}/src setup.py",
     "    docformatter --in-place --recursive {toxinidir}/src setup.py",
-    ]
+]
 use-flake8 = true
 
 [manifest]
@@ -110,7 +110,7 @@ additional-rules = [
     "recursive-include src *.xml",
     "recursive-include src *.zcml",
[...]

PR #215, opened 2023-10-07
Use tomlkit for TOML read/write operations

dataflake:
This PR is an alternative to #209. It does two things:

- replace `toml` use with `tomlkit`
- update dependencies in `requirements.txt` to the latest versions that still support Python 3.7

Running this code against current Zope yields the following diff in `.meta.toml`:

```diff
diff --git a/.meta.toml b/.meta.toml
index 82f0c085f..07cac2b24 100644
--- a/.meta.toml
+++ b/.meta.toml
@@ -2,7 +2,7 @@
 # https://github.com/zopefoundation/meta/tree/master/config/zope-product
 [meta]
 template = "zope-product"
-commit-id = "571024f0"
+commit-id = "babc889c"

 [python]
 with-pypy = false
@@ -11,6 +11,7 @@ with-sphinx-doctests = false
 with-windows = true
 with-macos = true
 with-future-python = true
+with-appveyor = false

 [coverage]
 fail-under = 80
```

The different is most likely due to the change in the `_set_python_config_value` method where I always use the default `False` because some values were `None` at that point, which `tomlkit` didn't like.


dataflake:
Instead of changing the behavior of `_set_python_config_value` I have decided to fix the real reason: Several script arguments that want to store true or false had an inconsistent default of `None`. I have corrected those. The outcome is still the same as the diff shown above.

sallner:
We can also do it this way, but it actually does not use the capabilities of `tomlkit` to add and preserve comments now. It does the same as before, so I am fine from this perspective.

sallner (commented):
It looks okay, to me, although we might end up with all default values in the `meta.toml` now.

dataflake:
> It looks okay to me, although we might end up with all default values in the `meta.toml` now.

I was thinking about that and to be honest I don't think that's a bad thing. It's very explicit.

icemac (approved):
Finally I got to this PR.

icemac:
Thank you for this PR. 😃

PR #216, opened 2023-10-07
Separate testenv setups for lint and package-checks

loechel:
combined tox testenvs for linting and package checks are not usefull, so seperate those two tests that has been together on a testenv:lint into testenv:package-checks and testenv:lint.

### DIFF
--- a/config/README.rst
+++ b/config/README.rst
@@ -271,7 +271,7 @@ updated. Example:
         "    src/foo/bar.py: E221 E222",
         "extend-ignore = D203, W503",
         ]
-    additional-addons = [
+    additional-plugins = [
         "maccabe"
         ]
     additional-sources = "testproj foo bar.py"
--- a/config/README.rst
+++ b/config/README.rst
@@ -271,6 +271,9 @@ updated. Example:
         "    src/foo/bar.py: E221 E222",
         "extend-ignore = D203, W503",
         ]
+    additional-addons = [
+        "maccabe"
+        ]
     additional-sources = "testproj foo bar.py"
 
     [manifest]
@@ -521,6 +524,10 @@ additional-config
   configuration section in ``setup.cfg``. *Caution:* This option has to be a
   list of strings so the leading white spaces and comments are preserved when
   writing the value to ``setup.cfg``.
+
+additional-plugins
+  Some packages want to have additional flake8 plugins installed.
+  *Caution:* This option has to be a list of strings.
 
 additional-sources
   Sometimes not only ``src`` and ``setup.py`` contain Python code to be checked
--- a/config/README.rst
+++ b/config/README.rst
@@ -98,7 +98,7 @@ Preparation
 
 The script needs a ``venv`` with some packages installed::
 
-    $ python3.11 -m venv .
+    $ python3 -m venv .
     $ bin/pip install -r requirements.txt
 
 To use the configuration provided here in a package call the following script::
--- a/config/default/tox-envlist.j2
+++ b/config/default/tox-envlist.j2
@@ -1,8 +1,8 @@
 [tox]
 minversion = 3.18
 envlist =
+    package-checks
     lint
-    py37
     py38
     py39
     py310
--- a/config/default/tox-envlist.j2
+++ b/config/default/tox-envlist.j2
@@ -3,6 +3,7 @@ minversion = 3.18
 envlist =
     package-checks
     lint
+    py37
     py38
     py39
     py310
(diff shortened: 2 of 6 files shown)


dataflake:
We all are aware that Python 3.7 is EOL. But dropping support for it requires a bit more discussion than just declaring it within this tool.

loechel:
@dataflake I can refactor that change out (Python 3.7 EOL) in a different pull request, as that needs to be done by a certain point. But I do want to have separate testenvs for linting and packaging checks

dataflake:
Thank you for factoring out the Python 3.7 parts. I'm ±0 on this PR - you're saying the combined lint environment is not useful, but don't provide an explanation why.

loechel:
Thanks @dataflake 

for me linting and checks if the package is releaseable is something different. 
I will run the lint step always before pushing any change to the github repo. 
But having unnecessary tests in the linting testenv that fails on every run (--> my case RestrictedPython) is frustrating. 
As RestrictedPython does not support future Pythons or even most recent without tests and adoption the lint has always failed. 

My own opinion is, that you should actually run all test always before commiting anything. in the long run I would like if we could get pre-commit in the default set of tools. But pre-commit should only test necessary test steps locally. a check of manfifest, or python-versions-check is not necessary locally.

dataflake:
I always run all tests myself before committing. I don't mind a test that always fails as long as I know why that is and there will be a fix coming at some point. I don't distinguish between "necessary" and other steps. So in the end I am still ±0 on this change. It seems to be a matter of taste.

dataflake:
Actually I just saw a redeeming factor. That new test builds the sdist and tests it. We didn't have that before, I like it.

dataflake (approved):

mauritsvanrees (approved):

config/README.rst
```diff
@@ -98,7 +98,7 @@ Preparation
 
 The script needs a ``venv`` with some packages installed::
 
-    $ python3.11 -m venv .
+    $ python3 -m venv .
```
icemac, line 101: This change is not correct: The code in this repository requires to use Python 3.11+.

config/README.rst
```diff
@@ -271,6 +271,9 @@ updated. Example:
         "    src/foo/bar.py: E221 E222",
         "extend-ignore = D203, W503",
         ]
+    additional-plugins = [
+        "maccabe"
```
icemac, line 275: I think you mean: `mccabe`.

config/README.rst
```diff
@@ -522,6 +525,10 @@ additional-config
   list of strings so the leading white spaces and comments are preserved when
   writing the value to ``setup.cfg``.
 
+additional-plugins
+  Some packages want to have additional flake8 plugins installed.
+  *Caution:* This option has to be a list of strings.
```
icemac, line 530: There is no need for extra `*Caution:*` as the option name is plural.

config/default/tox-envlist.j2
```diff
@@ -1,6 +1,7 @@
 [tox]
 minversion = 3.18
 envlist =
+    release-check
```
icemac, line 4: I think this env should also run on GHA but it should be allowed to fail without failing the build.
So the release manager is not surprised when running this env before the release that fixes are needed. (I personally rely on GHA being green before the release, I do not always run the tests localy before cutting a release.)

icemac (commented):
Sorry for being late to the party.

config/README.rst
dataflake: You changed that from Python 3.8 to 3.11 in https://github.com/zopefoundation/meta/commit/1ef8b9b8ae8311a7e79edb13d5086aed20bda382 but I can't see anything in that change set that requires 3.11+. I know that #209 would require 3.11, but that's not merged.

d-maurer:
Jens Vagelpohl wrote at 2023-10-17 03:28 -0700:
>@dataflake commented on this pull request.
>> @@ -98,7 +98,7 @@ Preparation
>
> The script needs a ``venv`` with some packages installed::
>
>-    $ python3.11 -m venv .
>+    $ python3 -m venv .
>
>You changed that from Python 3.8 to 3.11 in https://github.com/zopefoundation/meta/commit/1ef8b9b8ae8311a7e79edb13d5086aed20bda382 but I can't see anything in that change set that requires 3.11+. I know that #209 would require 3.11, but that's not merged.

Apparently, `tox` tries to access `tomllib`, available from
Python 3.11 on.

I got corresponding test failures with a Python 3.11 alpha release
and had to upgrade to a release version.

Issue #200, opened 2023-06-29
When to start dropping support for end-of-life Python versions?


icemac:
According to https://peps.python.org/pep-0537/#lifespan: "As of 2023-06-27, 3.7 has reached the [end-of-life phase](https://devguide.python.org/devcycle/#end-of-life-branches) of its release cycle."

So my question is: Can we come up with a decision/formula when to drop support for Python versions which are end of life? Should we begin to phase it out when it goes into end-of-live or wait some time (a month, half a year, a year, …)?

Any opinions?

dataflake:
I think this is one of those issues where everyone has their own policies, there is no "one size fits all" approach. My own personal opinion: I would drop it whenever it is convenient after EOL status has been reached. Convenient means whenever someone touches a package for some other reason. I wouldn't start yet another round of updates just for that alone.

In practical terms that means dropping it from the `meta` scripts and templates soon, and then any packages that gets updates will have it applied automatically.

dataflake:
P.S.: In this specific case we have it relatively easy: Plone 6 only supports Python 3.8-3.11, anyway, so there's a lot less people who would be surprised or affected in a negative way.

mauritsvanrees:
For the record, Plone 5.2 supported Python 3.7 until now, but we already stopped testing with it this month, and in July I will do a release where we officially drop support.
But we will try not to break it.  For example I definitely do **not** want to add `python_requires>=3.8`.  Anyway, any changes to zope packages are unlikely to affect Plone 5.2, as it uses Zope 4.

Question: when an individual package like `zope.interface` drops support for Python 3.7, will that be a major release?

For the Zope package itself, a major release would not make much sense: we would then have a major Zope release every year.  But Zope 5.8 already drops support for Python 3.6 in a minor release, so it makes sense to keep doing it that way.

dataflake:
I would also call a change in supported Python versions a major release for packages other than Zope. For Zope we already decided it's OK to do that in a minor release.

icemac:
Dropping support for a Python version next to the end of life of that version via `meta/config` means for me: When that package gets released (I support releasing it as a new major release.), all packages depending on it will automatically no longer support that Python version. So we should start dropping support at the leaves (i. e. Zope) to make it clear that individual packages no longer need to support that Python version.

Dropping support for a Python version means for me:
* running `pyupgrade` to upgrade the syntax to the new minimum Python version
[...]

PR #217, opened 2023-10-07
Drop support of Python 3.7

loechel:
Remove Python 3.7 from list of supported Python Versions, as of today (2023-10-07) Python 3.7 is already end of Life (since 2023-06-27). See https://devguide.python.org/versions/ for reference.

See also #200.


dataflake (approved):
I am fine with merging this PR because all it does is change the `drop-legacy-python` script. I don't know anyone who uses that script after Python 2 support has been removed...

icemac (changes requested):
This script has to be adapted to be useable to actually drop Python 3.7 support. It was already used on all zopefoundation packages to drop Python 2.7 to 3.6 support, so this part is no longer needed and the change log messages should not mention it any more.

But really dropping Python 3.7 support also means to adapt all the templates and remove 3.7 from tox.ini, the c code build scripts etc.

icemac:
Maybe the script should be reworked into one adding support for new version (aka 3.12) + the already existing dropping of outdated versions.

tseaver:
If for nothing else than a tiny contribution to reducing our carbon footprint, can we please quit doing CI on EOL Python versions as a matter of policy?  Under current rules, that would give us five "stable" Python versions to test, plus the one in alpha/beta for the succeeding release.

dataflake:
I think the time has come to rework the templates in order to get rid of Python 3.7. I'm going to extend this PR soon.

d-maurer (approved):

tseaver (approved):

dataflake:
@icemac I have removed references to the `drop-legacy-python.py` script from the Readme.

icemac:
> @icemac I have removed references to the `drop-legacy-python.py` script from the Readme.

@dataflake Do you envision to no longer to use `drop-legacy-python.py` or should it become part of the ususal update config run?

dataflake:
I don't use that script, for me the only use would be as a scripting example.

PR #218, opened 2023-10-11
Specify FUTURE_PYTHON_VERSION as version range

dataflake:
See https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#specifying-a-python-version and scroll down to `hyphen ranges`: It's apparently possible to specify a range of versions for a Python version to be used with the GitHub `setup-python` action. That way we won't have to update this package for every new prerelease. It also means testing for the value of `FUTURE_PYTHON_VERSION` to determine if `with_future_python` has to be set is no longer necessary. Only the package's `future-python` setting is considered.

This is not fully tested because there's no Python 3.13 preprelease yet.

### DIFF
--- a/config/config-package.py
+++ b/config/config-package.py
@@ -23,7 +23,7 @@ META_HINT_MARKDOWN = """\
 Generated from:
 https://github.com/zopefoundation/meta/tree/master/config/{config_type}
 --> """
-FUTURE_PYTHON_VERSION = ""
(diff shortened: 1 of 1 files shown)


mauritsvanrees (approved):
I tried it out in https://github.com/zopefoundation/zope.exceptions/pull/31 and the version ranges seem to work as intended.

PR #219, opened 2023-10-16
Revert "Specify FUTURE_PYTHON_VERSION as version range"

icemac:
Reverts zopefoundation/meta#218

Until there is a Python 3.13 alpha version at GHA the changes break running `config-package.py` if `with-future-python` is `true`. I think the merge was done too early.

We can revert this revert when the 3.13 alpha finally appears.

Needed for https://github.com/zopefoundation/zope.index/pull/45

### DIFF
--- a/config/config-package.py
+++ b/config/config-package.py
@@ -23,7 +23,7 @@ META_HINT_MARKDOWN = """\
 Generated from:
 https://github.com/zopefoundation/meta/tree/master/config/{config_type}
 --> """
-FUTURE_PYTHON_VERSION = "3.13.0-alpha - 3.13.0"
+FUTURE_PYTHON_VERSION = ""
 DEFAULT = object()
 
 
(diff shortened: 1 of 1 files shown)


dataflake:
I disagree. From all I can tell all that breaks is that tox and GHA cannot find the interpreter. Instead of reverting the change in this package it is easy to just switch the future python flag in the affected packages' `.meta.toml` and run the configure script again. You will have to re-run the configuration script again, anyway, no matter where you make the change.

dataflake:
I'll look at this again today.

icemac:
In https://github.com/python/pythoncapi-compat/pull/79 I learned that `3.13` as version specifier seems to be enough. According to https://github.com/python/pythoncapi-compat/actions/runs/6533893659/job/17739970573 a 3.13a1 version is already available.

It seems we can close this issue and replace the version specifier by `3.13` for future releases.

dataflake:
I can try that out later today.

icemac:
https://github.com/zopefoundation/DocumentTemplate/pull/73 shows that the version range you proposed also works.

dataflake:
https://github.com/zopefoundation/DocumentTemplate/actions/runs/6545745015/job/17774852634 proves that a simple "3.13" doesn't consider prereleases, the range as proposed is still the better solution.

dataflake (approved):
Since there is a 3.13 prerelease now I put the range specification back in.

icemac:
As there is now an alpha release, this PR is no longer necessary.

PR #223, opened 2023-11-14
Revert "Fix path to ``test`` script on Windows"

icemac:
Reverts zopefoundation/meta#157

Reason: In https://github.com/zopefoundation/DateTime/pull/57 windows tests break again and I found another solution there.


icemac:
This PR is wrong.

PR #239, opened 2024-05-13
A better way to specify a future Python version

### LINKED ISSUE #238: A better way to specify a Python prerelease version
Right now the value of ``FUTURE_PYTHON_VERSION`` in ``shared/packages.py`` is set as a range, currently ``3.13.0-alpha - 3.13.0``. That's a bit clunky because that version string is inserted in many places.

The documentation for the ``setup-python`` GH action mentions a cleaner way that uses the flag ``allow-prereleases: true``, so you could define ``FUTURE_PYTHON_VERSION`` as a "clean" ``3.13`` and write it all like this:

```
- uses: actions/setup-python@v4
  with:
    python-version: "3.13"
    allow-prereleases: true
```

See https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#allow-pre-releases.

dataflake:
Fixes #238 

Advantages are cleaner GHA configuration files and when the future Python version is finally released all releases will be picked up automatically, so packages don't need any immediate configuration changes.

This change will insert the flag ``allow-prereleases`` into other Python versions as well because I didn't see a way to test for the specific version on the Jinja side. The matrix is only interpreted on the GHA side during execution. But this is safe - a prerelease will only ever be selected if there is no normal release.


icemac (approved):
I like these changes!

PR #243, opened 2024-05-23
Pin setuptools again

dataflake:
Pin setuptools so...

- source distributions don't get dotted names replaced by underscores
- zc.buildout compatibility issues


icemac (approved):
Sorry, I already sprinkled some `setuptools < 69` onto `master` as I thought this PR is already merged.

Issue #181, opened 2022-12-23
Test failures on GHA macos runners


dataflake:
The `macos` GitHub Actions runners reveal several issues. Tests using Python 3.11 on both `macos-12` and `macos-11` runners as well as tests using Python 3.9 and 3.10 on `macos-12` runners fail during the buildout run's dependency installation step. The binary wheel for `zope.interface` is typically the first such package encountered, it is downloaded and then fails to install with a traceback similar to the one shown at the bottom.

These issues do not show up on any local macOS install with either a self-compiled Python or `homebrew`-installed Python that will by default compile to the specific machine architecture (`arm64` or `x86_64`). Later Python versions on the GHA runners, however, are compiled as `universal2`, which combines both architectures in "fat" binaries, using the `configure` flags `--enable-universalsdk --with-universal-archs=universal2`.

It seems neither `zc.buildout` nor `setuptools` deal with this correctly, and how we use `setuptools` with `setup_requires` to build binary packages that depend on other binary packages appears to be an issue as well.

- `zc.buildout` uses a completely unsuitable simplistic way to pick a package from a list of candidates returned from the PyPI index, see https://github.com/buildout/buildout/issues/609#issuecomment-1317190892. This algorithm cannot deal with situations where more than a single binary wheel is returned from PyPI, as would be the case when a wheel exists for the specific archtitecture *plus* a `universal2` wheel.
- `setuptools` appears to misinterpret the macOS version number baked into the wheel name, see https://github.com/pypa/setuptools/issues/3687

Some of this was already discussed in https://github.com/zopefoundation/Zope/pull/1071. I want to take this up again soon and hopefully find some solution.

```
root: Reading https://pypi.org/simple/zope.interface/
Getting distribution for 'zope.interface==5.5.2'.
root: Downloading https://files.pythonhosted.org/packages/ff/a1/788d02d891a5abdf842eec244b432891f56f5c23cb30504c35f70b588b85/zope.interface-5.5.2-cp39-cp39-macosx_10_9_x86_64.whl#sha256=40f4065745e2c2fa0dff0e7ccd7c166a8ac9748974f960cd39f63d2c19f9231f
While:
  Installing.
  Getting section alltests.
  Initializing section alltests.
  Installing recipe zc.recipe.testrunner.
  Getting distribution for 'zope.interface==5.5.2'.

An internal error occurred due to a bug in either zc.buildout or in a
recipe being used:
Traceback (most recent call last):
[...]

PR #248, opened 2024-05-29
Use custom test step for macOS

dataflake:
Workaround for the many problems with the universal2-based macOS GHA runner Python builds like #181 

With this change the default GHA test configuration uses a custom "Test" step for macOS runners that is hardcoded to call the respective tox environment and tacking on an environment marker. The tox configuration for each package can then react to that marker and e.g. pre-install dependencies like C-code wheels that zc.buildout cannot install itself.

Used/tested with Zope (https://github.com/zopefoundation/Zope/pull/1214)

### DIFF
--- a/config/default/tests.yml.j2
+++ b/config/default/tests.yml.j2
@@ -75,10 +75,6 @@ jobs:
           - { os: ["macos", "macos-latest"], config: ["3.9",   "docs"] }
   {% endif %}
           - { os: ["macos", "macos-latest"], config: ["3.9",   "coverage"] }
-          # macOS/Python 3.11+ is set up for universal2 architecture
-          # which causes build and package selection issues.
-          - { os: ["macos", "macos-latest"], config: ["3.11",  "py311"] }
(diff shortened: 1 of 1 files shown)


d-maurer (commented):
I would expect corresponding changes in (some) `tox.ini`s.

For non `Mac` envs we have support for `gha_test_environment` and `gha_test_commands`; for `Mac` there is not support for `gha_test_commands`.

dataflake:
The packages that need changes in their tox configuration will get them next time someone updates the configuration files because at that moment the tests will fail. I'm not going to go through all of them right now.

And yes, it would be nice to have macOS-specific settings for the `gha-`-keys, but for now I will wait until there's a real need for it because changes in a packages tox.ini are not sufficient.

icemac (approved):
I would like to have some documentation so we do not forget about this feature. It seems only be needed when buildout is envolved, which is only the case in the `zope-package` and `buildout-recipe` templates. All other use `pip` for the installation.

icemac:
Hm, I overlooked the enabled auto-merge. Maybe my suggestions still can get incorporated somehow.

PR #250, opened 2024-06-01
Prevent publishing wheels for unreleased Python versions

dataflake:
This change tries to prevent publishing wheels for future_python versions during the `manylinux` GHA step.

### DIFF
--- a/config/c-code/tests.yml.j2
+++ b/config/c-code/tests.yml.j2
@@ -416,7 +416,6 @@ jobs:
         run: sudo chown -R $(whoami) ${{ steps.pip-cache-default.outputs.dir }}
 {% if with_future_python %}
       - name: Prevent publishing wheels for unreleased Python versions
-        if: '%(future_python_version)s' != ''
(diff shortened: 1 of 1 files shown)


icemac (approved):

config/c-code/tests.yml.j2
```diff
@@ -414,6 +414,10 @@ jobs:
           name: manylinux_${{ matrix.image }}_wheels.zip
       - name: Restore pip cache permissions
         run: sudo chown -R $(whoami) ${{ steps.pip-cache-default.outputs.dir }}
+{% if with_future_python %}
+      - name: Prevent publishing wheels for unreleased Python versions
+        run: VER=$(echo '%(future_python_version)s') | tr -d .) && rm -f wheelhouse/*-cp${VER}*.whl
```
tseaver, line 419: The `.)` bit breaks with a bash syntax error:   it needs to be quoted, e.g. `tr -d ".)"`.

PR #251, opened 2024-06-01
Set with_future_python to false if no future python version is defined

dataflake:
This change overrides the value for `with_future_python` and sets it to `false` if there is no future Python version defined.

The goal is to "turn off" the flag with just a change in meta/config itself instead of having to touch all package configurations. So when a new Python version is released and no suitable next Python alpha/beta is available on GHA all you need to do is to set `shared.packages.FUTURE_PYTHON_VERSION` to an empty string and add the new Python release version to the various templates before re-running for a given package. The package won't need any changes itself. Once a new alpha/beta is on GHA that version number can be set again and the packages that do define `with_future_python` will pick it up the next time meta/config is run on them.

### DIFF
--- a/config/config-package.py
+++ b/config/config-package.py
@@ -229,9 +229,12 @@ class PackageConfiguration:
 
     @cached_property
     def with_future_python(self):
+        value = self._set_python_config_value('future-python')
+
         if not FUTURE_PYTHON_VERSION:
             return 'false'
-        return self._set_python_config_value('future-python')
+
+        return value
 
     @cached_property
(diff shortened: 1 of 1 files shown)


icemac:
According to https://peps.python.org/pep-0745/ 3.14a1 is to be released at 2024-10-15, this is 14 days after the planned final release of 3.13, see https://peps.python.org/pep-0719/.
So I'd expect there will be nearly always s future version we can work towards.

icemac (approved):

PR #255, opened 2024-06-14
Add script to update to currently supported Python versions.

icemac:
Remove previous script to drop support for old Python versions instead.

The script is now able to drop and add in one go and does not need to be adapted for version updates in the range of supported versions.

See https://github.com/zopefoundation/z3c.currency/pull/8 which was built using this PR.

### DIFF
--- a/config/README.rst
+++ b/config/README.rst
@@ -658,6 +658,22 @@ Hints
 * Call ``bin/check-python-versions <path-to-package> -h`` to see how to fix
   version mismatches in the *lint* tox environment.
 
+Updating to the currently supported Python versions
+---------------------------------------------------
+
+There is `update-python-support.py` which can be used to update a repository to
+the currently supported Python versions as defined in ``shared/package.py``.
+
+Usage
++++++
+
+To update a repository to the currently supported Python versions call::
+
+    $ bin/python update-python-support.py <path-to-package>
+
+It supports a parameter ``--interactive`` to not automatically commiting the
+changes.
+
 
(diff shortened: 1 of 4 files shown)


config/README.rst
```diff
@@ -658,6 +658,22 @@ Hints
...
+the currently supported Python versions as defined in ``shared/package.py``.
+
+Usage
++++++
+
+To update a repository to the currently supported Python versions call::
+
+    $ bin/python update-python-support.py <path-to-package>
+
+It supports a parameter ``--interactive`` to not automatically commiting the
```
dataflake, line 674:
```suggestion
It supports a parameter ``--interactive`` to not automatically commit the
```

dataflake: I'd really like an alternate spelling `--no-commit` for this flag so it is the same as for the main configure script, some way to make all changes but not do a `git commit`.

dataflake (changes requested):

dataflake:
The ``--no-commit`` flag is not all that different from ``--interactive`` because those zest.releaser scripts don't have a simple "don't commit" mode, they either commit everything without input or ask questions and then ask about a commit, which is annoying.

dataflake:
Tested with https://github.com/zopefoundation/MultiMapping/pull/9

icemac:
@dataflake I like the changes you made, so I enabled auto-merge which will fire when you will approve this PR.

dataflake (approved):

PR #258, opened 2024-06-27
fix: typo

tseaver:
Introduced in https://github.com/zopefoundation/meta/pull/250

### DIFF
--- a/config/c-code/tests.yml.j2
+++ b/config/c-code/tests.yml.j2
@@ -416,7 +416,7 @@ jobs:
         run: sudo chown -R $(whoami) ${{ steps.pip-cache-default.outputs.dir }}
 {% if with_future_python %}
       - name: Prevent publishing wheels for unreleased Python versions
-        run: VER=$(echo '%(future_python_version)s') | tr -d .) && rm -f wheelhouse/*-cp${VER}*.whl
+        run: VER=$(echo '%(future_python_version)s') | tr -d ".)" && rm -f wheelhouse/*-cp${VER}*.whl
 {% endif %}
       - name: Publish package to PyPI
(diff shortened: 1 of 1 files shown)


dataflake (approved):

dataflake:
That fix cannot be correct. You would quote the closing of the expression that starts at `$VER=$(`. The dot doesn't need to be quoted, either, it belongs to the `tr -d .` invocation. In #250 you refer to a bash syntax error without firther details, can you show that?

config/c-code/tests.yml.j2
```diff
@@ -416,7 +416,7 @@ jobs:
         run: sudo chown -R $(whoami) ${{ steps.pip-cache-default.outputs.dir }}
 {% if with_future_python %}
       - name: Prevent publishing wheels for unreleased Python versions
-        run: VER=$(echo '%(future_python_version)s') | tr -d .) && rm -f wheelhouse/*-cp${VER}*.whl
+        run: VER=$(echo '%(future_python_version)s') | tr -d ".)" && rm -f wheelhouse/*-cp${VER}*.whl
```
dataflake, line 419:
```suggestion
        run: VER=$(echo '%(future_python_version)s' | tr -d .) && rm -f wheelhouse/*-cp${VER}*.whl
```

dataflake (changes requested):
Can you try your test scenario with this change?

dataflake:
I committed my change and tested it on https://github.com/zopefoundation/zodbpickle/actions/runs/9708076471/job/26794281137. Looks like the wheelhouse folder is guarded by special permissions. I wonder if sudo can help.

dataflake (approved):

PR #259, opened 2024-06-28
feat: add 'isort_additional_config' option

tseaver:
FBO the forked stdlib modules in 'zodbpickle':  we want to keep the diff minimal between the original version and our fork, and so need to disable most kinds of linting entirely for them.

### DIFF
--- a/config/README.rst
+++ b/config/README.rst
@@ -558,6 +558,10 @@ additional-sources
   should be applied. This option has to be a string. It defaults to the empty
   string.
 
+additional-config
+  Additional options for the ``[isort]`` section.  This option has to be a
+  list of strings.
+
 GitHub Actions options
 ``````````````````````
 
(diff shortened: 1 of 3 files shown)


dataflake:
Suggestion: Rename this to "isort_additional_config". The naming convention isn't set firmly, but eyeballing the code I find it easier to see these prepended section names.

I don't see the supporting code in `config-package.py` and in the README, though.

tseaver:
@dataflake I have created PR #262 which attempts to rationalize the naming scheme for config options.  I will adjust this one accordingly after it merges.

tseaver:
@dataflake PTAL

I rebased the change after merging #262 to `master`, updating the names accordingly, and adding the `config/config-package.py` driver and `config/README.rst` docs.

dataflake (approved):

PR #270, opened 2024-08-27
Add primitive pyproject.toml file to silence a pip warning. Closes #266

### LINKED ISSUE #266: Pip deprecation warning on editable install
### What I did:

```shell
# create a venv somehow
$ pip install --editable 'git+https://github.com/zopefoundation/Zope.git@master#egg=Zope'
[…]
  DEPRECATION: Legacy editable install of Zope from git+https://github.com/zopefoundation/Zope.git@master#egg=Zope (setup.py develop) is deprecated. pip 25.0 will enforce this behaviour change. A possible replacement is to add a pyproject.toml or enable --use-pep517, and use setuptools >= 64. If the resulting installation is not behaving as expected, try using --config-settings editable_mode=compat. Please consult the setuptools documentation for more information. Discussion can be found at https://github.com/pypa/pip/issues/11457
[…]
$ python --version
Python 3.12.5
$ pip list | grep -E 'pip|wheel|setuptools'
pip                     24.2
setuptools              73.0.0
wheel                   0.44.0
$ sw_vers
ProductName:		macOS
ProductVersion:		14.6.1
BuildVersion:		23G93
```

### What I expect to happen:

This should install without warning

### What actually happened:

The warning that is displayed above. I'm not quite sure yet I understand this warning correctly, but what I read sounds like switching from setup.py to pyproject.toml could solve this?

### What version of Python and Zope/Addons I am using:

See above

This warning and the linked GitHub issue, seems to suggest that  pip 25 (soonish) might break editable installs of Zope.

Is that an assessment you guys share? If so, is this the point to switch to a pyproject.toml based workflow for Zope?

dwt:
Pip in version >= 24.2 generates a warning if a setup.py based project is installed with --editable that pip calling `setup.py develop` directly is deprecated and should be fixed by adding at least a basic pyproject.toml file. This file tells pip to use the setuptools entry point, thus fixing this (and likely more) warnings.

As I am not knowledgeable how this project works (sorry, I couldn't find any unit tests) I am not quite sure how to test these changes. I did try to apply the script to the Zope project, and it seemed to generate the `pyproject.toml` file.

However, I am not quite sure what this will do if such a file is already there. Please help. :)


dataflake:
This is not sufficient for a very simple reason: Some projects already have a `pyproject.toml` and the code as it stands would just overwrite it. The solution should use `tomlkit`, which is available in the configuration script, to attempt loading an existing file and then merge the new data into it before writing to disk.

dwt:
@dataflake how does testing this sort of things work in this project?

dataflake:
There are no tests. You make changes here and then run the configuration script against an existing packages to view the changes it makes there.

icemac:
I found only one usage of `pyproject.toml`: https://github.com/search?q=org%3Azopefoundation%20pyproject.toml&type=code
So it is probably not necessary to read in the possibly existing file but have configuration options in `.meta.toml` to configure what's in `pyproject.toml` and default to the content you currently have here.

mauritsvanrees:
In Plone we have those lines in all packages that use `plone/meta` (the Plone equivalent of `zopefoundation/meta`:

https://github.com/plone/meta/blob/main/config/default/pyproject.toml.j2#L1-L2

```
[build-system]
requires = ["setuptools>=68.2"]
```

Works fine.

dwt:
Now that we have #276 that is much more advanced than this pull request - shall I close it, or do you want to keep it around for comparison for a bit?

dataflake:
I'll just close it

PR #275, opened 2024-08-30
Add a tox environment that uses setuptools development code

dataflake:
This PR adds a new tox environment `setuptools-latest` that is not run by default and that inherits everything from the base testenv but overrides the dependencies to pull setuptools straight from the main branch on GitHub.

I tested it against Zope itself and `persistent` and nothing fails (yet), so other than eyeballing what ends up in the tox environment's site-packages I can't prove that it really helps. Yet.

### DIFF
--- a/config/default/tox-testenv.j2
+++ b/config/default/tox-testenv.j2
@@ -54,3 +54,11 @@ extras =
 {% for line in testenv_additional %}
 %(line)s
 {% endfor %}
+
+[testenv:setuptools-latest]
+basepython = python3
+deps =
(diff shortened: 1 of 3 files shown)


dataflake:
Another option would be to add a GitHub Actions run that tests with the latest setuptool against at least one current version of Python and either just Linux or even Windows if the project specifies Windows tests.

icemac (approved):
I like this idea, even though it requires manual work, but it does not result in douzens of emails when setuptools broke our workflows again.

dataflake:
Yes, that is a good point. GitHub will send emails for the scheduled test runs, so it it fails there will be a lot of them.

PR #276, opened 2024-09-01
Add ``pyproject.toml`` to the list of managed files

dataflake:
This PR is an alternative implementation of #270. It uses `tomlkit` to load an existing file and then merges some sane defaults before writing it to disk.

Successfully tested against Zope, which has no pre-existing `pyproject.toml`, and `BTrees`, which does.

@dwt: Can you check to see if this works for you and gets rid of the warnings?

### DIFF
--- a/config/README.rst
+++ b/config/README.rst
@@ -113,12 +113,13 @@ The script does the following steps:
    if it is not yet added.
 2. Copy ``setup.cfg``, ``tox.ini``, ``tests.yml``, ``MANIFEST.in``,
    ``.readthedocs.yaml`` (if needed), and ``.gitignore`` to the repository.
-3. Remove a possibly existing ``.coveragerc`` and ``bootstrap.py``. (Coverage
+3. Create or update a ``pyproject.toml`` project configuration file.
+4. Remove a possibly existing ``.coveragerc`` and ``bootstrap.py``. (Coverage
    is now configured in ``tox.ini`` for packages which are no buildout
    recipes.)
-4. Run the tests via: ``tox``. The ``tox`` script may be either on the current
+5. Run the tests via: ``tox``. The ``tox`` script may be either on the current
    ``$PATH`` or in the ``bin`` subfolder of the current working directory.
-5. Create a branch and a pull request. (Prevent an automatic commit of all
+6. Create a branch and a pull request. (Prevent an automatic commit of all
    changes with the command line switch ``--no-commit``, or an automatic push
    to GitHub using the command line switch ``--no-push``.)
 
--- a/config/README.rst
+++ b/config/README.rst
@@ -111,15 +111,15 @@ The script does the following steps:
 
 1. Add the package name to ``packages.txt`` of the selected configuration type
    if it is not yet added.
-2. Copy ``setup.cfg``, ``tox.ini``, ``tests.yml``, ``MANIFEST.in``,
+1. Copy ``setup.cfg``, ``tox.ini``, ``tests.yml``, ``MANIFEST.in``,
    ``.readthedocs.yaml`` (if needed), and ``.gitignore`` to the repository.
-3. Create or update a ``pyproject.toml`` project configuration file.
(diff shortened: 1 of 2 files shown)


config/config-package.py
```diff
@@ -550,6 +550,38 @@ def manifest_in(self):
...
+
+        buildsystem_cfg = pyproject_toml['build-system']
+
+        # Insert sane default for build-system:requires
+        requires = buildsystem_cfg.get('requires', [])
+        setuptools_requirement = [
+            x for x in requires if x.startswith('setuptools')]
+        if not setuptools_requirement:
+            requires.append(f'setuptools{SETUPTOOLS_VERSION_SPEC}')
+            buildsystem_cfg['requires'] = requires
```
icemac, line 571: How does it handle updating `SETUPTOOLS_VERSION_SPEC` to a newer version number?
If I understood the code correctly, it only writes the version spec once as long it does not exist.
Did I miss something here?

icemac (changes requested):

config/config-package.py
dataflake: You're correct, I did not take that into account. All I thought was "don't overwrite existing setuptools specs, the author will have had a good reason to put it there". I changed it to remove all setuptools specs it finds and just add our own.

Successfully tested against Zope and BTrees. Please re-check.

dwt:
Sorry, I only have time on Tuesdays to look into this stuff, so I hope I get to it tomorrow.

mauritsvanrees (approved):
LGTM.  I tested it by copying a pyproject.toml file from a plone project to a zope project, and then see what changes the script made.  It changed what had to be changed and left the rest alone.

config/config-package.py
```diff
@@ -679,6 +680,7 @@ def configure(self):
                 ".github/workflows/pre-commit.yml",
                 ".gitignore",
                 ".meta.toml",
+                "pyproject.toml",
```
icemac, line 683: As this file might not exist, it should be added to `early_add` in line 629, to make the linter tox run happy.

dataflake: Good catch, I move it.

config/README.rst
```diff
@@ -113,12 +113,13 @@ The script does the following steps:
...
+4. Remove a possibly existing ``.coveragerc`` and ``bootstrap.py``. (Coverage
    is now configured in ``tox.ini`` for packages which are no buildout
    recipes.)
-4. Run the tests via: ``tox``. The ``tox`` script may be either on the current
+5. Run the tests via: ``tox``. The ``tox`` script may be either on the current
    ``$PATH`` or in the ``bin`` subfolder of the current working directory.
-5. Create a branch and a pull request. (Prevent an automatic commit of all
+6. Create a branch and a pull request. (Prevent an automatic commit of all
    changes with the command line switch ``--no-commit``, or an automatic push
    to GitHub using the command line switch ``--no-push``.)
```
dwt, line 124: We could set the numbering everywhere to 1. here, as markdown will auto increment the numbers on rendering. That would lead to fewer diffs in the future, but at the cost of decreased readability on the shell.

dwt (approved):
Looks good to me. This is much more advanced than what I did last week, as it handles existing files well and should thus be easily usable even for projects outside of the Zope Foundation namespace.

Nice work.

config/README.rst
dataflake: That's a good idea but apparently not supported on GitHub's markdown implementation. I just tried and reverted it.

dwt:
oh... yeah, this is an RST file not a Markdown. Ah well.

dwt:
Seems to work well for me.

config/README.rst
mgedmin: ReStructuredText has auto-numbering too, if you use `#.` for the list items instead of actual numbers.

dataflake: Yes that works, thanks!

PR #279, opened 2024-09-15
Move Python 3.13 into the stable Python category and call a notional Python 3.14 "future Python"

dataflake:
This PR moves Python 3.13 into the list of normally tested and compatible Python versions and ends its status as future Python. Since 3.13.0.rc1 it is ABI-stable and there won't be any surprises, so we can start declaring full support where possible. I want to start doing the binary packages.

### DIFF
--- a/config/c-code/manylinux-install.sh.j2
+++ b/config/c-code/manylinux-install.sh.j2
@@ -30,13 +30,14 @@ yum -y install libffi-devel
 tox_env_map() {
     case $1 in
 {% if with_future_python %}
-        *"cp313"*) echo 'py313';;
+        *"cp314"*) echo 'py314';;
 {% endif %}
         *"cp38"*) echo 'py38';;
         *"cp39"*) echo 'py39';;
         *"cp310"*) echo 'py310';;
         *"cp311"*) echo 'py311';;
         *"cp312"*) echo 'py312';;
+        *"cp313"*) echo 'py313';;
         *) echo 'py';;
     esac
 }
@@ -45,15 +46,16 @@ tox_env_map() {
 for PYBIN in /opt/python/*/bin; do
     if \
 {% if with_future_python %}
-       [[ "${PYBIN}" == *"cp313/"* ]] || \
+       [[ "${PYBIN}" == *"cp314/"* ]] || \
 {% endif %}
        [[ "${PYBIN}" == *"cp311/"* ]] || \
        [[ "${PYBIN}" == *"cp312/"* ]] || \
+       [[ "${PYBIN}" == *"cp313/"* ]] || \
        [[ "${PYBIN}" == *"cp38/"* ]] || \
        [[ "${PYBIN}" == *"cp39/"* ]] || \
        [[ "${PYBIN}" == *"cp310/"* ]] ; then
 {% if with_future_python %}
-        if [[ "${PYBIN}" == *"cp313/"* ]] ; then
+        if [[ "${PYBIN}" == *"cp314/"* ]] ; then
             "${PYBIN}/pip" install --pre -e /io/
             "${PYBIN}/pip" wheel /io/ --pre -w wheelhouse/
         else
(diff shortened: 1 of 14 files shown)


config/c-code/manylinux-install.sh.j2
```diff
@@ -30,13 +30,14 @@ yum -y install libffi-devel
...
 {% if with_future_python %}
-        *"cp313"*) echo 'py313';;
+        *"cp314"*) echo 'py314';;
 {% endif %}
         *"cp38"*) echo 'py38';;
         *"cp39"*) echo 'py39';;
         *"cp310"*) echo 'py310';;
         *"cp311"*) echo 'py311';;
         *"cp312"*) echo 'py312';;
+        *"cp313"*) echo 'py313';;
```
icemac, line 40: Maybe it is a good point in time to compute these lines (and the other ones referring to Python versions) from the values set in `shared/packages.py`. This would make it easier to add/remove Python versions. What do you think?

config/c-code/tests-cache.j2
```diff
@@ -4,9 +4,7 @@
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-{% if with_future_python %}
           allow-prereleases: true
-{% endif %}
```
icemac, line 9: This means that e. g. `AccessControl` for all Python versions would be tested against the `dev` release for the future Python version instead of the latest stable release. I am not sure whether we always want this.

icemac (approved):
I like making Python 3.13 an officially supported version now and also the idea of starting to release packages with Python 3.13 support.

There are some suggestions regarding the PR.

d-maurer (approved):

config/c-code/tests-cache.j2
dataflake: The setting means it will only choose a prerelease if no stable release is available. It is required here because Python 3.13 has no final release. For all other Python versions it has no adverse effect.

config/c-code/manylinux-install.sh.j2
dataflake: IMHO it would be great to compute the selection of Python versions in all places where we want to say "all Python versions between the declared minimum and maximum", such as in `tox.ini` as well. I'd actually like to be in a place where I can more easily use zope.meta outside of zopefoundation packages, for example with work-related packages where I want to say "my minimum is Python 3.11".

dataflake: I have converted all hardcoded Python versions lists so they are generated by a loop.

config/c-code/tests-cache.j2
icemac: Cool, I did not know about this. Thank you for the clarification.

config/config-package.py
```diff
@@ -462,6 +472,8 @@ def tox(self):
             with_sphinx_doctests=self.with_sphinx_doctests,
             docs_deps=docs_deps,
             setuptools_version_spec=SETUPTOOLS_VERSION_SPEC,
+            future_python_shortversion=FUTURE_PYTHON_SHORTVERSION,
+            supported_python_versions=supported_python_versions(True),
```
icemac, line 476:
```suggestion
            supported_python_versions=supported_python_versions(short_version=True),
```

config/config-package.py
```diff
@@ -479,6 +491,8 @@ def tests_yml(self):
         gha_test_commands = self.gh_option('test-commands')
         require_cffi = self.meta_cfg.get(
             'c-code', {}).get('require-cffi', False)
+        py_version_matrix = [x for x in zip(supported_python_versions(False),
+                                            supported_python_versions(True))]
```
icemac, line 495:
```suggestion
        py_version_matrix = [
            x for x in zip(supported_python_versions(short_version=False),
                           supported_python_versions(short_version=True))]
```

icemac (approved):
I like it especially generating the lists of Python versions.

PR #281, opened 2024-09-25
Avoid generating an empty setenv =

mgedmin:
Because it breaks packages that have a

    test-additional = ["setenv =", ...]

in their .meta.toml, like zope.publisher.

### DIFF
--- a/config/default/tox-testenv.j2
+++ b/config/default/tox-testenv.j2
@@ -18,8 +18,8 @@ deps =
   {% if with_sphinx_doctests and with_future_python %}
     Sphinx
   {% endif %}
-setenv =
 {% if testenv_setenv %}
+setenv =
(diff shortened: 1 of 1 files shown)


dataflake:
Why wouldn't you take those `setenv` lines out of `test-additional` and add them to `testenv-setenv`?

mgedmin:
> Why wouldn't you take those setenv lines out of test-additional and add them to testenv-setenv?

Oh, I will.

But there's no reason to generate empty `setenv =` here anyway, is there?

dataflake:
There is no reason to generate an empty one, that is correct. I'm still worried that other things will break. Some of these replacements that get handed from template to sub-template are hard to follow. Go ahead and we'll see what happens.

dataflake (approved):

PR #282, opened 2024-09-27
Turn zope.meta into a real installable Python package

### LINKED ISSUE #97: RFE: packagize the metaconfiguration script

dataflake:
This PR turns zope.meta into a real Python package that can be installed like any other package. The scripts were changed slightly so they can be used as a console script entry point and the README was updated accordingly.

At the same time the packaging also allows running the zope.meta scripts against zope.meta itself, the results are also part of this PR.

My main goal is script re-use for custom applications. I would like to be able to create a simple script (or simple package) where I can import zope.meta code and override settings, like loading file templates from a different path or applying different settings from the zopefoundation defaults. This PR is just a first step towards that goal.

Fixes #97.


pyproject.toml
```diff
@@ -0,0 +1,21 @@
+# 
+# Generated from:
+# https://github.com/zopefoundation/meta/tree/master/config/c-code
```
mgedmin, line 3: This part is surprising to me, especially given the `template = "pure-python"` in .meta.toml.

CHANGES.rst
```diff
@@ -0,0 +1,5 @@
+Change log
+==========
+
+1.0 (unreleased)
+----------------
```
mgedmin, line 5: Maybe 

```suggestion
----------------
- Converted to a real installable package.
```

docs/hacking.rst
```diff
@@ -0,0 +1,212 @@
...
+===========================
+
+
+Getting the Code
+################
+
+The main repository for :mod:`zope.meta` is in the Zope Foundation
+Github repository:
+
+  https://github.com/zopefoundation/zope.meta
```
mgedmin, line 11: Do you plan to rename the repository after merging this?

I'm :+1: for that idea.

docs/hacking.rst
```diff
@@ -0,0 +1,212 @@
...
+To run the tests via ``zope.testrunner``:
+
+.. code-block:: sh
+
+   $ /tmp/hack-zope.meta/bin/zope-testrunner --test-path=src
+   Running zope.testrunner.layer.UnitTests tests:
+     Set up zope.testrunner.layer.UnitTests in 0.000 seconds.
+     Ran 41 tests with 0 failures, 0 errors and 0 skipped in 0.003 seconds.
+   Tearing down left over layers:
+     Tear down zope.testrunner.layer.UnitTests in 0.000 seconds.
```
mgedmin, line 72: Do we have tests for this code now?  That is wonderful news!

docs/hacking.rst
```diff
@@ -0,0 +1,212 @@
...
+
+   $ /tmp/hack-zope.meta/bin/coverage run -m zope.testrunner \
+      --test-path=src
+   ...
+   $ coverage report -i -m --fail-under=100
+   Name                                 Stmts   Miss Branch BrPart    Cover   Missing
+   ----------------------------------------------------------------------------------
+   src/zope/meta/__init__.py       4      0      0      0  100.00%
+   src/zope/meta/message.py       52      0     18      0  100.00%
+   src/zope/meta/tests.py        189      0     38      0  100.00%
```
mgedmin, line 87: This does not look truthful to me.  There's no message.py in src/zope/meta.

setup.py
```diff
@@ -0,0 +1,90 @@
...
+        'test': ['zope.testrunner'],
+        'docs': ['Sphinx', 'sphinx_rtd_theme'],
+    },
+    entry_points={
+        'console_scripts': [
+            'config-package=zope.meta.config_package:main',
+            'multi-call=zope.meta.multi_call:main',
+            're-enable-actions=zope.meta.re_enable_actions:main',
+            'set-branch-protection-rules='
+            'zope.meta.set_branch_protection_rules:main',
```
mgedmin, line 91:
```suggestion
            (
                'set-branch-protection-rules='
                'zope.meta.set_branch_protection_rules:main'
            ),
```

to make it more apparent that this is one list item split across two lines.

src/zope/meta/multi_call.py
```diff
@@ -38,24 +39,26 @@
 args, sub_args = parser.parse_known_args()
 packages = list_packages(args.packages_txt)
```
mgedmin, line 40: I think these two lines should be part of main() too.

mgedmin (approved):
Mostly LGTM!  Also, this PR makes me happy.

One script needs a bit more code to be moved inside main() so it won't do command-line argument parsing at run time.

The lack of actual tests makes me wonder about the utility of the tox.ini and docs/hacking.rst.  I think we should at least add an empty-ish tests.py so that anyone who wants to add a test will have a lower threshold of entry.  I don't think this PR needs to be blocked until we have actual tests -- as long as CI doesn't fail trying to enforce the 85% of coverage or whatever it has configured by default.

I would also like confirmation that all the references to github.com/zopefoundation/zope.meta are intentional and you plan to rename the repository.  I think GitHub will provide automatic redirection for people who use the old URL/have old git clones.

setup.py
```diff
@@ -56,6 +56,11 @@ def read(*rnames):
     ],
     license='ZPL 2.1',
     url='https://github.com/zopefoundation/zope.meta',
+    project_urls={
+        'Documentation': 'https://zopemeta.readthedocs.io',
+        'Issue Tracker': 'https://github.com/zopefoundation/meta/issues',
+        'Sources': 'https://github.com/zopefoundation/meta',
```
mgedmin, line 62: This is inconsistent with all the other places that already refer to https://github.com/zopefoundation/zope.meta

pyproject.toml
dataflake: The file was copied from another project, I fixed it.

CHANGES.rst
dataflake: I added a change log entry

docs/hacking.rst
dataflake: Those pages were copied from another package. I removed all test output, that was not meant to suggest real test output, just demonstrate the command invocation.

docs/hacking.rst
dataflake: No, it is sample output, not reality, and it's gone now.

setup.py
dataflake: Done

src/zope/meta/multi_call.py
dataflake: I moved all the argparse bits into the respective `main` function

setup.py
dataflake: That's all fixed.

d-maurer (approved):

docs/hacking.rst
```diff
@@ -0,0 +1,185 @@
...
+--------------------
+
+.. note::
+
+   Please ensure that all tests are passing before you submit your code.
+   If possible, your submission should include new tests for new features
+   or bug fixes, although it is possible that you may have tested your
+   new code by updating existing tests.
+
+If have made a change you would like to share, the best route is to fork
```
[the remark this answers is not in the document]
dataflake: Fixed

setup.py
```diff
@@ -0,0 +1,97 @@
...
+def read(*rnames):
+    with open(os.path.join(os.path.dirname(__file__), *rnames)) as stream:
+        return stream.read()
+
+
+setup(
+    name='zope.meta',
+    version='1.0.dev0',
+    author='Zope Foundation and Contributors',
+    author_email='zope-dev@zope.org',
```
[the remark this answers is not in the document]
dataflake: It still works, yes, but I just changed it to the zope.dev domain.

docs/hacking.rst
```diff
@@ -0,0 +1,185 @@
...
+the Githb repository, check out your fork, make your changes on a branch
+in your fork, and push it.  You can then submit a pull request from your
+branch:
```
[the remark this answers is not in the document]
dataflake: The entire documentation text was copied from another project and then adapted where needed. I did not write this myself.

I have changed that passage to mention the contributor status and to discourage private forks, though.

mauritsvanrees (approved):
I did not try it, but the idea seems good, making it easier to apply zope/meta on a package without needing to have a git checkout of it.

This would even make it possible to add a scheduled GitHub workflow action that installs `zope.meta` and runs it on the repo and see if anything has changed.

setup.py
```diff
@@ -0,0 +1,97 @@
...
+    keywords="zope packaging",
+    classifiers=[
+        'Development Status :: 3 - Alpha',
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: Zope Public License',
+        'Programming Language :: Python',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.8',
+        'Programming Language :: Python :: 3.9',
+        'Programming Language :: Python :: 3.10',
```
icemac, line 49: This makes my heart bleed a bit: Previously the minimum requiered version for meta/config was Python 3.11. And I think that I was using Python 3.11+ syntax somewhere. Being forced to write Python 3.8 code in the once most advanced repo feels kind of sad.
But I see the point that it is easier for self updating the repository to support all the versions we currently support for ZF projects.

icemac (approved):
I really like the idea and admire the time you invested into this change. I already used `meta/config` for some OSS projects outside the ZF and making reuse easier will be a great win.

Though I have some hopefully helpful questions.

icemac:
> This would even make it possible to add a scheduled GitHub workflow action that installs `zope.meta` and runs it on the repo and see if anything has changed.

You mean we could even make our repos self-updating. This is a really interesting idea. 😀

docs/_static/.gitignore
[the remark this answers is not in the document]
dataflake: It was so there's content in `docs/_static` and Sphinx would not warn about a missing folder. I removed it and commented out the `html_static_path` directive.

docs/hacking.rst
```diff
@@ -0,0 +1,192 @@
...
+its ``tox.ini`` file:
+
+- The ``lint`` environment runs various "code quality" tests on the source,
+  and fails on any errors they find.
+
+- The ``py38``, ``py39``, ``py310``, ``py311``, ``py312``, ``py313``, and
+  ``pypy3`` environments each build an environment from the corresponding
+  Python version, install :mod:`zope.meta` and testing dependencies,
+  and runs the tests.  It then installs ``Sphinx`` and runs the doctest
+  snippets.
```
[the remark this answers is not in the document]
dataflake: I removed the list of Python versions.

setup.py
dataflake: You were using `tomllib`, which (needlessly) introduced the hard Python 3.11 requirement. I replaced that with `tomlkit`, which is used everywhere else in the package. I find it more valuable to have a consistent set of supported versions everywhere. Less surprises.

dataflake: And don't forget, we could not even run the configure script against this package without that consistent Python version support because there's no functionality for a different set.

dataflake:
> I am not totally convinced about the renaming: In the archived packages we point to https://github.com/zopefoundation/meta to request unarchival. So maybe the code might go to zope.meta but the issues should stay in meta. Do you think this is reasonable and possible?

I am very surprised that you consider a potentially broken link in archived packages in any way important. I disagree and don't think that is important at all. Those are by definition outdated and abandoned.

I don't think it is reasonable and renaming a repository does not allow for keeping the issue tracker in the old place.

dataflake:
> > This would even make it possible to add a scheduled GitHub workflow action that installs `zope.meta` and runs it on the repo and see if anything has changed.
> 
> You mean we could even make our repos self-updating. This is a really interesting idea. 😀

I am scared of this. Please don't. That would mean even the smallest and safest commit now risks a broken package and extra work because some of those pre-commit hooks go red.

mauritsvanrees:
I suppose the repo could keep its `meta` name even when `setup.py` says `zope.meta`.

dataflake:
> I suppose the repo could keep its `meta` name even when `setup.py` says `zope.meta`.

That would also mean we have another strange one-off, in this case with a package name/repository name mismatch. I'd prefer avoiding that.

dataflake:
To finally get this off the ground I will retain the repository name and changed the documentation and configurations to make this work.

mauritsvanrees:
Thanks a lot!

icemac:
@dataflake Thank you for your work on this ticket, I really like the outcome!

PR #286, opened 2024-10-02
Support for specifying a minimum supported Python version

dataflake:
This PR adds support for a new command line parameter `--oldest-python` and a corresponding value in `.meta.toml` that allows setting a minimum supported Python version that may be different from the previously hardcoded default.

### DIFF
--- a/docs/narr.rst
+++ b/docs/narr.rst
@@ -176,6 +176,10 @@ The following options are only needed one time as their values are stored in
   a final release thus it is not yet generally supported by the zopefoundation
   packages.
 
+--oldest-python
+  The oldest version of Python supported by this package. Specified as version
+  number, e.g. ``3.8``.
+
 --with-docs
   Enable building the documentation using Sphinx. This will also create a
   configuration file `.readthedocs.yaml` for integration with
--- a/docs/narr.rst
+++ b/docs/narr.rst
@@ -178,7 +178,8 @@ The following options are only needed one time as their values are stored in
 
 --oldest-python
   The oldest version of Python supported by this package. Specified as version
-  number, e.g. ``3.8``.
+  number, e.g. ``3.8``. This setting is optional and defaults to the lowest
+  Python version generally supported by zopefoundation packages.
 
 --with-docs
   Enable building the documentation using Sphinx. This will also create a
--- a/src/zope/meta/config_package.py
+++ b/src/zope/meta/config_package.py
@@ -110,9 +110,8 @@ def handle_command_line_arguments():
     parser.add_argument(
         '--oldest-python',
         dest='oldest_python',
-        default=OLDEST_PYTHON_VERSION,
-        help='Oldest supported Python version. Defaults to'
-             f'Python {OLDEST_PYTHON_VERSION}.')
+        help='Oldest supported Python version. Defaults to:'
+             f' {OLDEST_PYTHON_VERSION}.')
     parser.add_argument(
         '--with-docs',
         # people (me) use --with-sphinx and accidentally
@@ -207,8 +206,9 @@ class PackageConfiguration:
 
     @cached_property
     def oldest_python(self):
-        value = (self.meta_cfg['python'].get('oldest-python') or
-                 self.args.oldest_python)
+        value = (self.args.oldest_python or
+                 self.meta_cfg['python'].get('oldest-python') or
+                 OLDEST_PYTHON_VERSION)
         try:
             version = parse_version(value)
         except InvalidVersion:
--- a/src/zope/meta/config_package.py
+++ b/src/zope/meta/config_package.py
@@ -112,7 +112,7 @@ def handle_command_line_arguments():
         dest='oldest_python',
         default=OLDEST_PYTHON_VERSION,
         help='Oldest supported Python version. Defaults to'
-             f'Python {OLDEST_PYTHON_VERSION}')
+             f'Python {OLDEST_PYTHON_VERSION}.')
     parser.add_argument(
         '--with-docs',
         # people (me) use --with-sphinx and accidentally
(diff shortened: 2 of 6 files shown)


docs/narr.rst
```diff
@@ -176,6 +176,11 @@ The following options are only needed one time as their values are stored in
   a final release thus it is not yet generally supported by the zopefoundation
   packages.
 
+--oldest-python
+  The oldest version of Python supported by this package. Specified as version
+  number, e.g. ``3.8``. This setting is optional and defaults to the lowest
+  Python version generally supported by zopefoundation packages.
+
```
d-maurer, line 183: Should we also have a symmetric `newest-python`?

src/zope/meta/config_package.py
```diff
@@ -105,6 +107,12 @@ def handle_command_line_arguments():
         default=False,
         help='Activate support for a future non-final Python version if not'
         ' already configured in .meta.toml.')
+    parser.add_argument(
+        '--oldest-python',
+        dest='oldest_python',
+        default=OLDEST_PYTHON_VERSION,
+        help='Oldest supported Python version. Defaults to'
+             f'Python {OLDEST_PYTHON_VERSION}.')
```
d-maurer, line 115:
```suggestion
        help='Oldest supported Python version. Defaults to: '
             f'{OLDEST_PYTHON_VERSION}.')
```
This way, the help text gives an indication on the value syntax.

src/zope/meta/config_package.py
```diff
@@ -197,6 +205,21 @@ def config_type(self):
                 'Please use `--type` to select it.')
         return value
 
+    @cached_property
+    def oldest_python(self):
+        value = (self.meta_cfg['python'].get('oldest-python') or
+                 self.args.oldest_python)
```
d-maurer, line 211: I assume that `self.args` refers to the command line arguments. Usually, those take precedence over other configuration sources.

src/zope/meta/config_package.py
```diff
@@ -348,7 +371,7 @@ def pre_commit_config_yaml(self):
             "pre-commit-config.yaml.j2",
             self.path / ".pre-commit-config.yaml",
             self.config_type,
-            oldest_python_version=OLDEST_PYTHON_VERSION.replace(".", ""),
+            oldest_python_version=self.oldest_python.replace(".", ""),
```
d-maurer, line 374: Will this break should we ever get a version component number >= 10?

d-maurer (approved):

src/zope/meta/config_package.py
mgedmin: I've adopted a habit of putting the spaces on the beginning of each line, this way they're all aligned and it's really noticeable when you forget to add one:

```python
    long_string_literal=(
        "long text goes brrrrrrrr"
        " continuation line 1"
        " continuation line 2"
        " continuation line 3"
    ),
```

src/zope/meta/update_python_support.py
```diff
@@ -107,21 +113,23 @@ def main():
...
 
         if not_yet_supported:
-            for version in sorted(list(not_yet_supported)):
-                call(
-                    bin_dir / 'addchangelogentry',
-                    f'Add support for Python {version}.',
-                    *non_interactive_params)
-            python_versions_args = ['--add=' +
-                                    ','.join(supported_python_versions())]
+            version_spec = ', '.join(sorted(not_yet_supported))
```
mgedmin, line 124: Does this properly sort 3.9 before 3.10?

(I suppose it doesn't matter; I don't think there are any packages left where we would be _adding_ 3.9 support.)

mgedmin (approved):
LGTM

docs/narr.rst
dataflake: You're welcome to suggest a patch.

src/zope/meta/config_package.py
dataflake: I don't understand this comment. This line removes dots. It will work the same with a version number `10.5` as it does with `3.8`. Or am I misunderstanding something?

src/zope/meta/update_python_support.py
dataflake: Not it does not properly sort and it did not properly sort before my changes. I (and probably Michael as the original author as well) didn't think it worthwhile to add the complication of parsing all these strings into proper version specifications, sorting them and then turning them back into strings just so they are sorted nicer in a change log entry. That is the only place where you would see it.

src/zope/meta/config_package.py
dataflake: Applied both changes

src/zope/meta/config_package.py
dataflake: You're right, I changed the ordering.

d-maurer:
Jens Vagelpohl wrote at 2024-10-2 23:33 -0700:
>@dataflake commented on this pull request.
>
>
>
>> @@ -348,7 +371,7 @@ def pre_commit_config_yaml(self):
>             "pre-commit-config.yaml.j2",
>             self.path / ".pre-commit-config.yaml",
>             self.config_type,
>-            oldest_python_version=OLDEST_PYTHON_VERSION.replace(".", ""),
>+            oldest_python_version=self.oldest_python.replace(".", ""),
>
>I don't understand this comment. This line removes dots. It will work the same with a version number `10.5` as it does with `3.8`. Or am I misunderstanding something?

Removing the dots will lose information: `3.12` and `31.2`
are mapped to the same `312`.

dataflake:
> Removing the dots will lose information: `3.12` and `31.2` are mapped to the same `312`.

Sorry Dieter, that is a bridge we can worry about when Python 10 arrives. Both of us may no longer be around.

PR #287, opened 2024-10-15
Fix branch protection min version.

icemac:
It has to to start with `py` like `NEWEST_PYTHON`.

Used in https://github.com/zopefoundation/zc.recipe.testrunner/pull/18

### DIFF
--- a/.meta.toml
+++ b/.meta.toml
@@ -2,7 +2,7 @@
 # https://github.com/zopefoundation/meta/tree/master/config/pure-python
 [meta]
 template = "pure-python"
-commit-id = "35d10997"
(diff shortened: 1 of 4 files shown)


icemac:
Coverage looks strange: It seems https://github.com/zopefoundation/meta/commit/210c1a0c5b3340f18c0bc7e54682c3d9a55a0673 dropped the coverage below 20 % but I do not understand why.

dataflake (approved):

dataflake:
Honestly, that one test is nothing but a placeholder right now to enable testing for coverage at all. Just dial down the percentage. It does not matter.

icemac:

PR #291, opened 2024-12-17
Retire Python 3.8

dataflake:
This PR sets Python 3.9 as the oldest supported Python version and it removes Python 3.8 support from this package as well.

- Python 3.8 has been EOL for more than two months, see https://devguide.python.org/versions/
- The upcoming Plone 6.1 sets an even higher minimum supported version (3.10)
- zc.buildout really wants to drop it as well, see https://github.com/buildout/buildout/pull/671
- In Zope the list of custom version pins to keep Python 3.8 support gets longer and longer, this is an unnecessary maintenance burden.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,7 +21,6 @@ jobs:
         config:
         # [Python version, tox env]
         - ["3.11", "release-check"]
-        - ["3.8", "py38"]
         - ["3.9", "py39"]
         - ["3.10", "py310"]
         - ["3.11", "py311"]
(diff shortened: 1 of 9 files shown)


d-maurer (approved):

icemac:
@dataflake O no, you were way quicker than me and I walked top to bottom through my inbox. I'll merge my PR into yours.

icemac (approved):
I like it!

mauritsvanrees:
@dataflake Does this mean you would be fine with `zc.buildout` dropping Python 3.8 support?  The upcoming 4.0.0 would be a good time for that.

PR #292, opened 2024-12-18
Allow `setuptools <= 75.6.0`.

icemac:
`zc.buildout` is nowadays tested against this version. Maybe we addidtionally need a final release of `zc.buildout` as support for these newer versions is only in `4.0.0a1`.


icemac:
See also https://github.com/buildout/buildout/issues/673

icemac:
Hm, that `setuptools` version does not support Python 3.8 anymore. So I'd also suggest starting to drop support for that Python version. (I'd like to do this in a sepate PR.)

icemac:
> Hm, that `setuptools` version does not support Python 3.8 anymore. So I'd also suggest starting to drop support for that Python version. (I'd like to do this in a sepate PR.)

See https://github.com/zopefoundation/meta/pull/293

dataflake:
> `zc.buildout` is nowadays tested against this version. Maybe we addidtionally need a final release of `zc.buildout` as support for these newer versions is only in `4.0.0a1`.

I think it's fine. https://github.com/buildout/buildout/pull/671 didn't make any code changes to zc.buildout so it should work with the current release.

dataflake (approved):

icemac:

PR #303, opened 2025-02-11
Add script to convert package to PEP 420 namespace.

icemac:
Used for https://github.com/zopefoundation/hurry.query/pull/18

### DIFF
--- a/src/zope/meta/pep_420.py
+++ b/src/zope/meta/pep_420.py
@@ -84,24 +84,22 @@ def main():
 
         setup_py = []
         for line in (path / 'setup.py').read_text().splitlines():
-            if 'find_packages' in line:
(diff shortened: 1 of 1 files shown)


icemac:
I simplified the implementation a bit more because @mauritsvanrees pointed me to an easier way which even sould prevent a warning during the release if there are directories which do not contain any Python code.

icemac:

mauritsvanrees:
Since all packages in the zopefoundation org have a `src` layout, we could also use [automatic package discovery](https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#automatic-discovery).

That means these lines (plus the import of `find_namespace_packages`) could be removed:

```
    packages=find_namespace_packages('src'),
    package_dir={'': 'src'},
```

PR #313, opened 2025-03-06
Update to `setuptools <= 75.8.2` and `zc.buildout >= 4.1.3`.

icemac:
Used in https://github.com/zopefoundation/groktoolkit/pull/77 and https://github.com/zopefoundation/zopetoolkit/pull/139


icemac:
@dataflake I updateted the PR according to to your suggestions and created https://github.com/zopefoundation/groktoolkit/pull/78 from it which is also green.

CHANGES.rst
```diff
@@ -18,6 +18,8 @@ Change log
 
 - Add ``pyupgrade-exclude`` to ``[pre-commit]`` section in ``.meta.toml``.
 
+- Update to ``zc.buildout < 4.1``.
```
dataflake, line 21:
```suggestion
- Downgrade ``zc.buildout`` version pin to ``<4.1`` due to
  ongoing setuptools compatibility issues.
```

dataflake (approved):
I would clearly spell out a reason for that pin, just so we know.

PR #348, opened 2025-10-02
Allow setuptools less than 81.

### LINKED ISSUE #337: `setuptools == 78.1.1` might be too strict
## QUESTION

We are currently setting `setuptools == 78.1.1` in the `[build-system]` section of `pyproject.toml`. This might be necessary for packages installed using `zc.buildout` but not for all types of packages.

On the other hand: Is `pyproject.toml` even read by `zc.buildout` based installations?

Maybe we are too strict with this restriction as we are already lagging behind 2 major setuptools version (latest version is 80.9.0 as of writing this).

And I saw broken tests in several packages which were still using `setuptools <= 75.6.0` after releasing `zope.event` using setuptools 78. This means that the constraint can also create the same kind of problems (aka broken tests/builds) we want to prevent with it.

Current example of us user running into a problem with the constraint: https://github.com/zopefoundation/zope.annotation/issues/28

dataflake:
Closes #337


dataflake:
I have now changed it to `>= 80.9.0,< 81`, which works. Tested against Zope itself.

dataflake:
I've set it to `>= 78.1.1,< 81` and that works for Zope itself.

mauritsvanrees (approved):

PR #365, opened 2025-11-24
Use one centralized definition for commonly used script arguments

dataflake:
Many of the scripts here use a common set of script arguments. Instead of copying and pasting the same argument definitions in every script file I put it in a central place so scripts can just import this shared argument parser definition and then add their own custom arguments to it before parsing.


icemac (approved):
Nice improvement!

PR #368, opened 2025-11-25
Add script to move package metadata from setup.py to pyproject.toml

dataflake:
This script automates moving metadata from the `setup.py` invocation of `setup` to `pyproject.toml`.

This is a messy procedure. It does run through about 80 zopefoundation packages without breaking, though. As the documentation explains, the user must eyeball the changes.

### DIFF
--- a/src/zope/meta/setup_to_pyproject.py
+++ b/src/zope/meta/setup_to_pyproject.py
@@ -26,6 +26,7 @@ import tomlkit
 from .shared.call import call
 from .shared.git import git_branch
 from .shared.packages import META_HINT
+from .shared.packages import OLDEST_PYTHON_VERSION
 from .shared.packages import get_pyproject_toml
 from .shared.path import change_dir
 from .shared.script_args import get_shared_parser
@@ -110,7 +111,7 @@ def parse_setup_function(ast_node, assigned_names=None):
     return setup_kwargs
 
 
-def setup_args_to_toml_dict(setup_kwargs):
+def setup_args_to_toml_dict(setup_py_path, setup_kwargs):
     """ Iterate over setup_kwargs and generate a dictionary of values suitable
     for pyproject.toml and a dictionary with unconverted arguments
     """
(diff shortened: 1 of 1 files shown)


src/zope/meta/setup_to_pyproject.py
```diff
@@ -0,0 +1,448 @@
...
+    if not path.is_dir():
+        print(f' - {path} is not a folder')
+        sane = False
+
+    if not (path / 'setup.py').exists():
+        print(' - no setup.py found, cannot convert package.')
+        sane = False
+
+    if not (path / '.meta.toml').exists():
+        print(' - no pyproject.toml found, cannot convert package.')
```
icemac:
```suggestion
    if not (path / '.meta.toml').exists():
        print(' - no .meta.toml found, cannot convert package.')
```

src/zope/meta/setup_to_pyproject.py
```diff
@@ -0,0 +1,448 @@
...
+    # Create the ast tree for the setup module to find the setup call
+    # definition in order to parse out the call arguments.
+    ast_tree = ast.parse(file_contents)
+    setup_node = None
+
+    for ast_node in ast_tree.body:
+        if isinstance(ast_node, ast.Expr) and \
+           isinstance(ast_node.value, ast.Call) and \
+           ast_node.value.func.id == 'setup':
+            setup_node = ast_node.value
```
icemac:
```suggestion
            setup_node = ast_node.value
            break
```

src/zope/meta/setup_to_pyproject.py
```diff
@@ -0,0 +1,448 @@
...
+    license.replace('ZPL 2.1', 'ZPL-2.1')
+    p_data['license'] = license
+
+    classifiers = setup_kwargs.pop('classifiers', [])
+    new_classifiers = []
+    for classifier in classifiers:
+        if classifier.startswith('License'):
+            continue
+        elif classifier == 'Framework :: Zope2':
+            new_classifiers.append('Framework :: Zope :: 2')
```
icemac: I'd suggest to omit this classifier: I doubt that any of our packages is still compatible with Zope 2.

src/zope/meta/setup_to_pyproject.py
```diff
@@ -0,0 +1,448 @@
...
+                                                      'zope-dev@zope.dev'))
+        elif 'Zope Corporation' in classifier:
+            new_classifiers.append(classifier.replace('Zope Corporation',
+                                                      'Zope Foundation'))
+        else:
+            new_classifiers.append(classifier)
+
+    p_data['classifiers'] = new_classifiers
+
+    p_data['readme'] = 'README.rst'
```
icemac: Should we check that this file actually exists? (There might be some packages where it is still called `README.txt` – we could require to manually convert it here.)

src/zope/meta/setup_to_pyproject.py
```diff
@@ -0,0 +1,448 @@
...
+        branch_name = args.branch_name or "convert-setup-py-to-pyproject-toml"
+        updating = git_branch(branch_name)
+
+        if args.commit:
+            if args.commit_msg:
+                commit_msg = args.commit_msg
+            else:
+                commit_msg = ('Move package metadata from setup.py'
+                              ' to pyproject.toml.')
+            call('git', 'commit', '-m', commit_msg)
```
icemac: Might we also add a change log entry like in https://github.com/zopefoundation/meta/blob/24cb1653095b1806e55bb9d9968ce848de801a10/src/zope/meta/pep_420.py#L47-L51?

src/zope/meta/setup_to_pyproject.py
dataflake: The code now leaves out both of those styles for Zope 2.

src/zope/meta/setup_to_pyproject.py
```diff
@@ -0,0 +1,448 @@
...
+        elif classifier == 'Framework :: Zope2':
+            new_classifiers.append('Framework :: Zope :: 2')
+        elif classifier == 'Framework :: Zope3':
+            new_classifiers.append('Framework :: Zope :: 3')
+        elif 'zope-dev@zope.org' in classifier:
+            new_classifiers.append(classifier.replace('zope-dev@zope.org',
+                                                      'zope-dev@zope.dev'))
+        elif 'Zope Corporation' in classifier:
+            new_classifiers.append(classifier.replace('Zope Corporation',
+                                                      'Zope Foundation'))
```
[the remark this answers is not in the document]
dataflake: I was applying the replacements from `setup.py` in the wrong place. I moved them.

PR #369, opened 2025-11-26
Add option to show file changes after a failed ``tox`` ``lint`` run or not

dataflake:
One of the default behaviors that always bothered me is the fact that a failed ``lint`` run would always dump a diff with all uncommitted changes to the console. This may be fine if you have a habit of committing very small changes, but when you have made a lot of changes (most of which will be totally unrelated to any linting failure) and all of a sudden the screen scrolls away with a huge diff then that's seriously unhelpful bordering on annoying. I always have to scroll back to see the actual linting failure messages.

This PR adds an option to toggle that diff on or off. The default if it's unset is still on, so there is no behavior change for existing packages unless you explicitly want it.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Change log
 2.2 (unreleased)
 ----------------
 
+- Add option to show file changes after a failed ``tox`` ``lint`` run or not.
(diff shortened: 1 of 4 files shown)


icemac (approved):
LGTM.

PR #370, opened 2025-11-26
Supercharge template and variable override mechanism

dataflake:
I use the overrides option quite often and constantly find myself mistyping it, like typing `--override-templates`. This PR renames the option to just `--overrides`, not just to prevent typos but also because I extended it.

If you put a file `overrides.toml` into the `--overrides` folder its values will automatically override variable values with the same name inside `shared/packages.py`. This is useful for several of them when you use zope.meta to manage packages outside of the zopefoundation GitHub organization.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,9 @@ Change log
 
 - Rename the ``--template-overrides`` option to simply ``--overrides``.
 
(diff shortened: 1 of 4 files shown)


dataflake:
I renamed the PR and extended the description because I expanded the PR.

PR #407, opened 2026-04-13
Add Windows ARM64 support to c-code CI template

Harishmcw:
This PR updates the c-code CI template to support Windows ARM64 (windows-11-arm) runners.

#### tests-strategy.j2
- Added `windows-11-arm` to the OS matrix when `with_windows` is enabled
- Added exclude for `pypy` on `windows-11-arm` (not supported)
- Added exclude for oldest Python version on `windows-11-arm` (not supported)

#### tests-download.j2
- Split into two separate download steps for Linux/macOS and Windows
- Windows step uses `${{ runner.arch }}` in artifact name to distinguish between `windows-latest` (X64) and `windows-11-arm` (ARM64)

#### tests-download-linux.j2 (new file)
- Single download step for jobs that only run on Linux (docs, release-check)

#### tests.yml.j2
- Updated `docs` and `release-check` jobs to use `tests-download-linux.j2`
- Split Windows and Linux upload steps separately with `runner.arch` in Windows artifact name

## Related
- Closes zopefoundation/zope.interface#366

- [x] I signed and returned the [Zope Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the zopefoundation GitHub organization.

### DIFF
--- a/src/zope/meta/c-code/tests-strategy.j2
+++ b/src/zope/meta/c-code/tests-strategy.j2
@@ -12,16 +12,22 @@
           - "%(future_python_version)s"
 {% endif %}
 {% if with_windows %}
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest, windows-11-arm]
 {% else %}
         os: [ubuntu-latest, macos-latest]
 {% endif %}
-{% if with_pypy or gha_additional_exclude %}
+{% if with_pypy or gha_additional_exclude or with_windows %}
         exclude:
 {% endif %}
 {% if with_pypy %}
           - os: macos-latest
             python-version: "pypy-%(pypy_version)s"
+          - os: windows-11-arm
+            python-version: "pypy-%(pypy_version)s"
+{% endif %}
+{% if with_windows %}
+          - os: windows-11-arm
+            python-version: "3.10"
 {% endif %}
 {% for line in gha_additional_exclude %}
           %(line)s
--- a/src/zope/meta/c-code/tests.yml.j2
+++ b/src/zope/meta/c-code/tests.yml.j2
@@ -160,7 +160,7 @@ jobs:
         run: |
           ls -l dist
           twine check dist/*
-{% for kind in ('macOS x86_64', 'macOS arm64', 'all other platforms') %}
+{% for kind in ('macOS x86_64', 'macOS arm64') %}
       - name: Upload %(package_name)s wheel (%(kind)s)
         if: >
   {% if kind == 'macOS x86_64' %}
@@ -168,8 +168,6 @@ jobs:
   {% elif kind == 'macOS arm64' or kind == 'macOS universal2' %}
           startsWith(runner.os, 'Mac')
           && !startsWith(matrix.python-version, 'pypy')
-  {% else %}
-          !startsWith(runner.os, 'Mac')
   {% endif %}
         uses: actions/upload-artifact@v7
         with:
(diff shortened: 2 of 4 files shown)


src/zope/meta/c-code/tests-strategy.j2
```diff
@@ -12,16 +12,22 @@
...
 {% endif %}
 {% if with_pypy %}
           - os: macos-latest
             python-version: "pypy-%(pypy_version)s"
+          - os: windows-11-arm
+            python-version: "pypy-%(pypy_version)s"
+{% endif %}
+{% if with_windows %}
+          - os: windows-11-arm
+            python-version: "%(oldest_python_version)s"
```
dataflake: I find this condition problematic. The release designated as oldest supported Python version will shift and it makes no sense to exclude any oldest Python version like this. The check should be more specific and only apply to versions that are truly unsupported.

dataflake (changes requested):

src/zope/meta/c-code/tests-strategy.j2
Harishmcw: Thank you for the feedback!

I've updated the exclude to hardcode python-version: "3.10" specifically, as that is the only Python version unsupported on Windows ARM64,

src/zope/meta/c-code/tests.yml.j2
```diff
@@ -192,10 +190,21 @@ jobs:
...
           path: dist/*universal2.whl
-  {% else %}
-          path: dist/*whl
   {% endif %}
 {% endfor %}
+      - name: Upload %(package_name)s wheel (Windows)
+        if: startsWith(runner.os, 'Windows')
+        uses: actions/upload-artifact@v4
+        with:
+          name: %(package_name)s-${{ runner.os }}-${{ matrix.python-version }}-${{ runner.arch }}.whl
```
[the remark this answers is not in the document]
Harishmcw: I have updated Windows/Linux upload steps to use `actions/upload-artifact@v7`.

dataflake (approved):

Harishmcw:
Hi,

Thanks for merging the PR.

Could you please let me know when the workflows in [zope.interface](https://github.com/zopefoundation/zope.interface) will be regenerated using the updated template so that the ARM64 jobs run in CI?

I’m looking forward to using the WoA wheel via PyPI.

Thanks!

dataflake:
Could it be that you have not tested applying your changes to an actual package that uses `zope.meta`, like e.g. `zope.interface`, to make sure it works on the GitHub test runners? For an unrelated `zope.meta` task I did apply all latest `zope.meta` changes to `zope.interface` and I see strange issues that did not exist before:

First, there are intermittent failures in the `build-package` step building Windows/arm packages, like for 3.13/windows-11-arm on https://github.com/zopefoundation/zope.interface/actions/runs/24452665495/job/71448084541 right now. There's an issue with the `pip cache` create/restore operations. The specific Python version is not relevant, I have seen this failure on two different ones so far when re-running all tests.

When I re-run the tests and tell it to only re-run failing tests the broken `build-package` succeeds. But when I click on the Windows `test` steps and look at the output for `pip cache` I see bad cache keys, like `Windows-pip-3.15` for `test (3.14, windows-11-arm)`.

It seems there is at least an issue setting up cache keys correctly for use with the `actions/cache` GitHub action now that you have two separate Windows runs for one and the same Python interpreter. So far, setting up these cache keys relied on a simple two element combination of Python interpreter version and OS name. This doesn't work when you add the CPU platform as third element. Please take a look at that.

dataflake:
I have created #409 to track this issue.

PR #410, opened 2026-04-16
Fix pip cache key to include architecture for Windows ARM64 support

### LINKED ISSUE #409: Windows ARM64 c-code template changes issues
The support for building Windows ARM64 packages was added to the c-code templates in #407. It seems to cause issues:

First, there are intermittent failures in the build-package step building Windows/arm packages, like for 3.13/windows-11-arm on https://github.com/zopefoundation/zope.interface/actions/runs/24452665495/job/71448084541 right now. There's an issue with the pip cache create/restore operations. The specific Python version is not relevant, I have seen this failure on two different ones so far when re-running all tests.

When I re-run the tests and tell it to only re-run failing tests the broken build-package succeeds. But when I click on the Windows test steps and look at the output for pip cache I see bad cache keys, like Windows-pip-3.15 for test (3.14, windows-11-arm).

It seems there is at least an issue setting up cache keys correctly for use with the actions/cache GitHub action now that there are two separate Windows runs for one and the same Python interpreter. So far, setting up these cache keys relied on a simple two element combination of Python interpreter version and OS name. This doesn't work when you add the CPU platform as third element.

Harishmcw:
This PR updates the cache key logic in the c-code GitHub Actions template to include the runner architecture.

Previously, cache keys were based only on the OS and Python version:

`${{ runner.os }}-pip-${{ matrix.python-version }}`

With the addition of Windows ARM64 jobs, this caused cache collisions between x64 and ARM64 builds for the same Python version, leading to intermittent failures and incorrect cache restores.

Changes

- Updated cache key to include runner architecture: `${{ runner.os }}-${{ runner.arch }}-pip-${{ matrix.python-version }}`

- Updated restore keys to include architecture as well

This prevents cache collisions between x64 and ARM64 jobs and avoids incorrect cache reuse, stabilizing CI runs.

- [x] I signed and returned the [Zope Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the zopefoundation GitHub organization.

Closes #409

### DIFF
--- a/src/zope/meta/c-code/tests-cache.j2
+++ b/src/zope/meta/c-code/tests-cache.j2
@@ -32,7 +32,7 @@
           path: ${{ steps.pip-cache-default.outputs.dir }}
           key: %(cache_key)s
           restore-keys: |
-            ${{ runner.os }}-pip-
+            ${{ runner.os }}-${{ runner.arch }}-pip-
 
       - name: pip cache (Windows)
         uses: actions/cache@v5
@@ -41,4 +41,4 @@
           path: ${{ steps.pip-cache-windows.outputs.dir }}
           key: %(cache_key)s
           restore-keys: |
-            ${{ runner.os }}-pip-
+            ${{ runner.os }}-${{ runner.arch }}-pip-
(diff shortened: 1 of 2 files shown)


dataflake (approved):
Thank you for the quick fix, my test with zope.interface works now, see https://github.com/zopefoundation/zope.interface/actions/runs/24494382074/job/71586313269

PR #416, opened 2026-04-30
Move zest.releaser config from setup.cfg to pyproject.toml

icemac:
The `zest.releaser` configuration options do not work when set in `setup.cfg`. This PR moves them to the `[tool.zest-releaser]` section in `pyproject.toml` instead.

### Changes
- Removed zest.releaser section from `setup.cfg.j2` template
- Added zest.releaser config writing to `pyproject_toml()` in `config_package.py`
- For `c-code` packages, `create-wheel` and `upload-pypi` are set as proper TOML booleans (`false`)
- Updated documentation in `docs/narr.rst`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Change log
 2.2 (unreleased)
 ----------------
 
+- Move ``zest.releaser`` configuration from ``setup.cfg`` to
+  ``pyproject.toml`` where it is the ``[tool.zest-releaser]`` section.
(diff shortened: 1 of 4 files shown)


icemac:
These are the changes I was talking about to ease using `zest.releaser` for c-code repositories preventing an accidental upload of the sdist while releasing and instead rely on trusted publishing via GHA.

dataflake (approved):

PR #417, opened 2026-05-01
Drop cache action `restore-keys` to prevent bad cache hits

dataflake:
As observed in several places like https://github.com/zopefoundation/zope.security/pull/128 the `restore-keys` setting can lead to wrong caches used.

The line break at the end of the file is on purpose to reduce visual clutter in the resulting tests configuration.


icemac (approved):
Yes, let's get rid of the restore keys I am not sure why we had them in the first place. Might be helpful in Python only project but even there probably not always.

dataflake:
I bet it was simply copy-pasted from example code in the cache action documentation.

PR #418, opened 2026-05-06
Add optional publishing to PyPI with Trusted Publishing

dataflake:
This PR adds a new `.meta.toml` section `[pypi]` and the option `trusted-publishing` to enable publishing wheel and sdist for non-`c-code` projects when a tag is pushed to GitHub.

Tested on https://github.com/zopefoundation/Missing/pull/17

Switching a package from manual to trusted publishing involves additional configuration work on the package's configuration on pypi.org and GitHub. That's why this is currently purely optional and defaults to `false` where not specified.


dataflake:
Sorry for the large diff, while I was in the documentation files to add a page explaining how to switch a package to trusted publishing I have split another documentation file up so they are easier to maintain.

docs/narr.rst
```diff
@@ -664,6 +667,22 @@ manylinux-aarch64-tests
...
+https://pypi.org/.
+
+trusted-publishing
+  Explicitly enable or disable automated building and publishing of release
+  packages to PyPI using the Trusted Publishing process when a tag is pushed
+  to GitHub. Defaults to ``false``.
+  This option has no meaning for ``c-code`` template projects, those always
+  build and publish release packages using Trusted Publishing.
+  See https://docs.pypi.org/trusted-publishers/ for details on the Trusted
+  Publishing process.
```
icemac: Does it require any preparations on PyPI side to be able to switch it to true?

icemac: As I read further down the PR: the answer is yes, could you add a hint here that there is more documentation and necessary preparation explained there?

icemac (commented):
I like these changes. Besides the comment and the change request in the Missing PR: LGTM

docs/narr.rst
dataflake: I added a documentation link and also code to disable zest-releaser build/upload options, just like for c-code packages

icemac (approved):
LGTM.

PR #419, opened 2026-05-08
Update branch protection rules in ``set_branch_protection_rules``

dataflake:
This PR tweaks the branch protection rules codified in the helper script ``set_branch_protection_rules``. The rules we have are decent already, maybe the fact that they have been enforced through the script for a long time now has helped us duck the attacks we see on the Collective repositories.

Each of these changes is meant as a starting point for discussion, none of it is set in stone:

- Enforce "Dismiss stale pull request approvals when new commits are pushed": This results in a little more work in some cases where a PR is changed after reviewer approval. Most of the time reviewers will not approve when they want to see changes, so this is an edge case. But it prevents unreviewed commits on the protected branch.
- Add the organization group "Release Managers" to the setting "Allow specified actors to bypass required pull requests ". I am not sure about this one, it seems it's no longer necessary after adding the organization-wide "All-repository admin" role to the group to make the standard development workflow work again for non-admins.
- Enforce "Require branches to be up to date before merging", which is not really security-related but seems a good idea, anyway.

I also update the GitHub API version to be used from the old `2022-11-28` to the most recent `2026-03-10`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ Change log
 2.2 (unreleased)
 ----------------
 
+- Update the branch protection rules for ``set_branch_protection_rules``
+
 - Add an option to enable releases to be built and published automatically
   using PyPI Trusted Publishing for non-``c-code`` projects.
 
--- a/src/zope/meta/set_branch_protection_rules.py
+++ b/src/zope/meta/set_branch_protection_rules.py
@@ -34,7 +34,7 @@ def _call_gh(
         'gh', 'api',
         '--method', method,
         '-H', 'Accept: application/vnd.github+json',
-        '-H', 'X-GitHub-Api-Version: 2022-11-28',
+        '-H', 'X-GitHub-Api-Version: 2026-03-10',
         f'/repos/{ORG}/{repo}/branches/{DEFAULT_BRANCH}/{path}',
         *args, capture_output=capture_output,
         allowed_return_codes=allowed_return_codes)
@@ -45,7 +45,7 @@ def set_branch_protection(
     result = _call_gh(
         'GET', 'protection/required_pull_request_reviews', repo,
         allowed_return_codes=(0, 1))
-    required_pull_request_reviews = None
+    required_pr_reviews = None
     if result.returncode == 1:
         if json.loads(result.stdout)['message'] != "Branch not protected":
             # If there is no branch protection we create it later on using the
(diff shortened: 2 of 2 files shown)


icemac:
Let me reply here instead on the code:

> * Enforce "Dismiss stale pull request approvals when new commits are pushed": This results in a little more work in some cases where a PR is changed after reviewer approval. Most of the time reviewers will not approve when they want to see changes, so this is an edge case. But it prevents unreviewed commits on the protected branch.

Yes, this could be seen as a security feature but it makes live harder than necessary. I've seen this feature activated at a customer, requiring re-review for each and every change. (There it was combined with "Require branches to be up to date before merging", so even each merge from `master` required a new review – this was a nightmare for development speed.)

So I'd suggest not to enable this feature but instead maybe enforce only secure 2FA factors so the chance of taking over an account is minimized. On the other hand trust the members to not do anything malicious after a positive review.

> * Add the organization group "Release Managers" to the setting "Allow specified actors to bypass required pull requests ". I am not sure about this one, it seems it's no longer necessary after adding the organization-wide "All-repository admin" role to the group to make the standard development workflow work again for non-admins.

I am not sure whether we need it. If not, I'd prefer less configuration in the repository if possible.

> * Enforce "Require branches to be up to date before merging", which is not really security-related but seems a good idea, anyway.

I see no added value for this enforcement: I've nearly never seen any problem with merging a PR which did not have merge conflicts but was not up to date with the parent branch. So I'd prefer not to set it.

> I also update the GitHub API version to be used from the old `2022-11-28` to the most recent `2026-03-10`.

I was not aware that the API version is set via a date, but I like the update.

dataflake:
Thank you for the comments.

I only partially agree with your opinion about dismissing reviews after additional commits and forcing the branch to be up to date with its parent: I totally see your point in the situation you describe, a customer where development speed is important. But this is not what we have here. We have a very large set of repositories where the only development is done by a handful of people and consists mostly of package management-related changed. There is hardly any active development and speed is (mostly) not important. The main job it seems is maintenance, and making sure the repositories are not compromised by malicious actors.

I do agree that the branch protection rule setting "Allow specified actors to bypass required pull requests" for Release Managers is most likely pointless now. It did not really work before and in response the Release Managers team now has organization-wide "all-repository admin", which should solve the problem. I'm going to go through and remove that rule exception where we have it.

I'm going to close this PR because there isn't anything useful left here. The requirement for secure 2FA is now enforced.

Just FYI, I also reconfigured how teams are assigned repository permissions. I switched from assigning them in each and every repository to assigning them organization-wide at the top and removing all these individual team assignments at the repository level. That simplifies administration and makes custom role assignments at the repository level easier to spot during auditing. I eyeballed all those repository-level assignments and confirmed that they did the exact same as the organization-level permission assignment now.

PR #429, opened 2026-06-30
Stop testing Python 3.15 for now

icemac:
The `comrak` librarary cannot be installed, see
https://github.com/zopefoundation/meta/actions/runs/28322442376/job/83906624439


dataflake:
I think it is wrong to just give up testing for Python 3.15 just because zest.releaser has a badly broken dependency chain that pulls in full Rust installs. That's madness. This feels like a zest.releaser issue and I'd rather concentrate on that problematic pin:

```
hint: `comrak` (v0.0.12) was included because `zest-releaser` (v9.9.1) depends
  on `readme-renderer[md]` (v45.0) which depends on `comrak`
```

dataflake:
Suggestion for an alternative fix at #430

PR #430, opened 2026-06-30
Restrict readme-renderer for internal tests

dataflake:
Alternative fix to #429 

Restrict ``readme-renderer`` dependency to versions older than 45 for the internal tests. This restriction can be removed when ``readme-renderer`` and its dependencies install cleanly under Python 3.15.


icemac (approved):
Okay, let's try this idea.

PR #434, opened 2026-08-05
Configure zest.releaser to not add extra message with trusted publishing

mauritsvanrees:
Otherwise, with the usual extra message `[ci skip]`, no CI will run, so the trusted publishing code is never called. See https://github.com/zopefoundation/Products.PluggableAuthService/pull/127#issuecomment-5193860157


dataflake (approved):
I'll trust you that it works - I don't use zest.releaser.

PR #442, opened 2026-08-13
Fix disappearing top comment in ``pyproject.toml`` with newer ``tomlkit``

dataflake:
Fixes #440 

Manually tested with ``tomlkit`` latest release (0.15.1) and the previously working release 0.14.0.

This code will cause duplicate comments under these circumstances:
- the sandbox has ``tomlkit`` 0.14.0 or earlier AND
- the package to be configured has the older-style comment at the top of ``pyproject.toml``

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@ Change log
 2.2 (unreleased)
 ----------------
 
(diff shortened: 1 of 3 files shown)


icemac (approved):
Could it be an idea to require `tomlkit >= 0.15` to cirvumvent one of the possible stumbling stones?

